PR #5, opened 2015-06-29
D1506 add dfly patch

zrj-rimwis:
Extra patch target.
Couldn't detect old/new bsd.ports.mk so offset will be slightly off.

### DIFF
--- a/ports/devel/tbb/Makefile.DragonFly
+++ b/ports/devel/tbb/Makefile.DragonFly
@@ -1,4 +1,4 @@
-pre-patch:
+dfly-patch:
 	@${REINPLACE_CMD} -e 's/__FreeBSD__/(__FreeBSD__||__DragonFly__)/g' \
 		${WRKSRC}/include/tbb/tbb_config.h 			\
 		${WRKSRC}/include/tbb/machine/mic_common.h		\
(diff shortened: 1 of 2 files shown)


jrmarino:
i don't understand this "Couldn't detect old/new bsd.ports.mk so offset will be slightly off."

To create a new a new diff, one takes /usr/ports/Mk/bsd.port.mk, applies patch < bsd.port.mk.diff, makes changes to changed bsd.port.mk, then regenerate bsd.port.mk.diff (with genpatch)

zrj-rimwis:
I tried to apply current bsd.port.mk.diff on top freebsd-ports version, but there are several hunks that differ from dports bsd.port.mk version.
Maybe other patch is applied later? Or i have newer version in my today's fetch from freebsd-ports.

jrmarino:
hmm, the current diff applied fine to the freebsd bsd.port.mk 22 hours ago.

zrj-rimwis:
ah, iirc different hunks are these:
https://github.com/freebsd/freebsd-ports/commit/1bf8b3f2f5ed58e00a1b0ecd06b6fc8135318736#diff-f085c6a25e29526d6c30744e779cd602

jrmarino:
I didn't have any trouble at all. the existing bsd.port.mk.diff applies with lots of offsets, but it applies without error.  I just recreated the diff with the change: d05f5bc616ad7bb97b12e3dd595cc2e085a4ccd1

jrmarino:
manually added second change too, so closing PR without merge.
59da27defa86942ac4192e3088636b3210a62642

PR #6, opened 2015-07-01
D1507 fix pack tcl tk

zrj-rimwis:
simple changes for compilation of Tk/Tcl

### DIFF
--- a/ports/astro/tkgeomap/Makefile.DragonFly
+++ b/ports/astro/tkgeomap/Makefile.DragonFly
@@ -1,0 +1,2 @@
+dfly-patch:
+	${REINPLACE_CMD} -e 's/FreeBSD-\*/FreeBSD-*|DragonFly-*/g' ${WRKSRC}/configure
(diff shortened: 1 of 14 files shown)


jrmarino:
My laptop drive crashed 1 day before the whole laptop was supposed to be replaced (was already set for data transfer).  I lost access to my private key so I can't build these for a day or so.

jrmarino:
tclbsd failed:
cd /wrkdirs/devel/tclbsd/work/tclbsd-1.6 ; autoconf # needed to gen configure first
/bin/sh: autoconf: not found
*** Error code 127

yajl-tcl failed:
===>  Patching for yajl-tcl-1.6
cd /wrkdirs/devel/yajl-tcl/work/yajl-tcl-1.6 ; autoconf # needed to gen configure first
/bin/sh: autoconf: not found
*** Error code 127

zrj-rimwis:
for some reason USE_AUTOTOOLS= autoconf doesn't work or i missunderstood how to invoke autotools

zrj-rimwis:
ah i think i got the idea
it works if i change target: dfly-patch -> run-autotools-fixup but that is a bit too hack-ish

jrmarino:
i assume the problem is that you invoke autotools, but autotools is not a listed dependency.  (thus not available within poudriere jail)

jrmarino:
i didn't look though, I just assumed that was the cause of the issue.

PR #7, opened 2015-07-01
D1507 fix graphics

zrj-rimwis:
Fix for flightgear included
multimedia/avidemux-plugins fails to build with mp4 format support

### DIFF
--- a/ports/devel/codeblocks/Makefile.DragonFly
+++ b/ports/devel/codeblocks/Makefile.DragonFly
@@ -1,0 +1,6 @@
+USE_CXXSTD=	c++0x
+USES+=	alias
+
+dfly-patch:
+	${REINPLACE_CMD} -e 's@!(defined(__clang__) && __has_feature(cxx_nullptr))@ __cplusplus < 201103L @' \
+		${WRKSRC}/src/include/prep.h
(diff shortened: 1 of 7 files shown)


jrmarino:
gstream1-plugins-gl has about 80 errors, similar to:

jrmarino:
In file included from ../../../gst-libs/gst/gl/gstgl_fwd.h:26:0,
                 from gl.h:29,
                 from gstgldisplay.c:56:
../../../gst-libs/gst/gl/glprototypes/base.h:52:23: error: unknown type name 'GLenum'
                      (GLenum target, GLuint texture))

zrj-rimwis:
for some reason it doesn't pick up substitution in multimedia/gstreamer1-plugins/Makefile.Dragonfly
but it builds fine for me locally in dports

jrmarino:
check the makefile for ".include" e.g. MASTERDIR or .CURDIR/../../

if that is present, then it's pulling in the Makefile.DragonFly from that location.
In these cases, it's sometimes better just to use a diff, but each case is unique.

zrj-rimwis:
cannot reproduce breakage locally
easiest way would be to upstream typo fix to gstreamer1 git repo

jrmarino:
The problem was just as I guessed it was.
So was the fix: ff76fa45f04a597fc93b0247add8e95d176f519f

PR #8, opened 2015-07-01
D1507 fix guile2

zrj-rimwis:
with these changes even devel/autogen (5.18.4)
compiles and only have a single FAIL (date tag after "It has been AutoGen-ed")

### DIFF
--- a/ports/devel/boehm-gc/dragonfly/patch-dyn_load.c
+++ b/ports/devel/boehm-gc/dragonfly/patch-dyn_load.c
@@ -1,0 +1,11 @@
+--- dyn_load.c.orig	2014-06-03 09:08:02.000000000 +0300
++++ dyn_load.c
+@@ -435,7 +435,7 @@ GC_INNER GC_bool GC_register_main_static
+ # pragma weak dl_iterate_phdr
+ #endif
+ 
+-#if (defined(FREEBSD) && __FreeBSD__ >= 7)
++#if (defined(FREEBSD) && (defined(__DragonFly__) || __FreeBSD__ >= 7))
+   /* On the FreeBSD system, any target system at major version 7 shall   */
+   /* have dl_iterate_phdr; therefore, we need not make it weak as above. */
+ # define HAVE_DL_ITERATE_PHDR
--- a/ports/devel/boehm-gc/dragonfly/patch-tools_threadlibs.c
+++ b/ports/devel/boehm-gc/dragonfly/patch-tools_threadlibs.c
@@ -1,0 +1,11 @@
+--- tools/threadlibs.c.orig	2014-06-03 09:08:02.000000000 +0300
++++ tools/threadlibs.c
+@@ -42,7 +42,7 @@ int main(void)
+ #       ifdef GC_USE_DLOPEN_WRAP
+           printf("-ldl ");
+ #       endif
+-#       if (__FREEBSD_version >= 500000)
++#       if (__FREEBSD_version >= 500000) || defined(__DragonFly__)
+           printf("-lpthread\n");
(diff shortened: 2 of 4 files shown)


jrmarino:
i tried several times to fix guile2.  thank you!

jrmarino:
too good to be true.  guile2 fails just like it always has:
.libs/libguile_2.0_la-posix.o: In function `scm_tmpnam':
posix.c:(.text+0x1df9): warning: warning: tmpnam() possibly used unsafely; consider using mkstemp()
  CCLD     guile
./.libs/libguile-2.0.so: warning: warning: tmpnam() possibly used unsafely; consider using mkstemp()
  GEN      guile-procedures.texi
Segmentation fault (core dumped)
Makefile:3732: recipe for target 'guile-procedures.texi' failed

jrmarino:
I think this is a bug in boehm-gc:
++#if (defined(FREEBSD) && (defined(__DragonFly__) || __FreeBSD__ >= 7))

it should be:
#if defined(__DragonFly__) || ( (defined(FREEBSD) && __FreeBSD__ >= 7) )

zrj-rimwis:
defined(FREEBSD) should be used as is.
Have you rerun devel/boehm-gc compile first? libunistring too
It looks it uses previous version of those.

zrj-rimwis:
guile2 depends on threaded version (devel/boehm-gc-threaded) and it wasn't rebuilt
boehm-gc-7.4.2_3.log                               02-Jul-2015 13:58               48887
boehm-gc-redirect-7.4.2_3.log                      26-Jun-2015 07:02               48575
boehm-gc-threaded-7.4.2_3.log                      25-Jun-2015 00:38
for me it build fine just now after reruning on cleaned out pkg

jrmarino:
i wasn't told the order was important so I did it in reverse (e.g. guile first)

jrmarino:
okay, guile2 builds and is in dports now.

jrmarino:
Hmm, for some reason it says my comment is your comment.  Weird bug.

jrmarino:
repeating to work around bug:

"defined(FREEBSD) should be used as is."

I am not agreeing with this. at best this is confusing as hell

ORIGINAL translates to "IF FreeBSD release 7 or higher"

NEW translates to: "True if one of the folllowing is true:
1) FREEBSD AND DRAGONFLY at the same time (IMPOSSIBLE without external cflags)
2) FREEBSD release 7 or higher

This condition will NEVER be true for DragonFly unless the FREEBSD definition somehow includes DragonFly

I really need this spelled out
1) Why it's correct
2) Why is what I suggested not correct

If somehow they are both correct, I still want to use my suggestion. It's more clear.

zrj-rimwis:
it is not a bug, i did it intentionally, i guess it could be shifted around if it looks better
only thing that matters to me are only results of "gmake check" in build directory:
w/o patch-dyn_smth.c : 4 FAIL test guaranteed
w/ patch: 0 or 1 (depending if you are lucky) iirc random failures of gctest

since FREEBSD is a gc-boehm platform type (unless some one implements and pushes DRAGONFLY upstream) technically it should be checked too (not to be confused with \__DragonFly\__ os type)

i have some debugging notes but will not be able to access those until Tuesday
(currently travelling)
p.s. this is the main reason why i prefer iterative "distributed polishing" approach to single "private repolish" one

jrmarino:
I don't really like grouping 8 ports together.
I could like 7 and not like 1, but I have to choose to take all or nothing.

If I take all, then I have to fix the one I have questions about.
I don't know how hard making 1 PR per port is though.

I would have questioned this before importing it (regardless of ultimate decision)

PR #9, opened 2015-07-02
D1507 fix blender

zrj-rimwis:
allows to compile few more ports
blender will be useful for drm debugging


jrmarino:
So I'll wait for an update or a new pull request, right?

zrj-rimwis:
hmm didn't expected that these games are played here too.. very SVN like workflow.
I usually go eric cartman on first signs of -pedantic-errors and/or NiH/RtW syndromes.

this is just a first order make build deps patch (/proc version is working) that only needs small 2nd order tuning.
graphics/blender has huge dependencies tree and i'm more concerned about satisfying dependencies first then making sure program outputs its full path instead a.out for diagnostics reasons if user has not mounted /proc.
OpenBSD version certainly works without this anyway.
What i really concerned is how to work around missing malloc_usable_size() that breaks compile in current ports version of blender and locking bug in drm/i915 that blender easily reproduces.

I'll be unavailable until next Tuesday (parallel decomposition code refactoring weekend in one of our fortran models).
i guess being pedantic on fixing never imported ports in one iteration is more important than helping to fix kernel panic in drm/i915
So this can be closed due to -pedantic-errors && submitter is -EBUSY, pending next bulk build.

jrmarino:
"games"?

All I see is a peer review saying the fix isn't the correct one.  This is completely normal.

Secondly, using Deltaports is the backup or temporary fix.  Whenever possible and practical, the port has to be fixed at the FreeBSD ports level.  In the case that we are moving fixes from dports to port, the fixes *have* to be correct otherwise they'll just get rejected there.

Thirdly, there has already been leeway based on a port that doesn't have best fix is still better than no port at all.

I don't see any need to rush here.  Nobody is being pedantic.  When you have time, then resubmit.

jrmarino:
by the way, experience has shown that nobody comes back and improves patches on a working port.  Thus, you have to fix it correctly the first time because that's the last time anyone will look at it until it breaks again.

PR #11, opened 2015-07-09
fixes openjdk8 networking issues on DFly4.2

seanwh1te:
fix is slightly more complex than that for openjdk7 due to the jdk build
changes
  http://lists.dragonflybsd.org/pipermail/users/2015-January/207360.html

the change to the build file was taken from here:
  http://hg.openjdk.java.net/bsd-port/jdk8/jdk/rev/ffe5580b349c
so it will probably go away someday


jrmarino:
Okay, this will be incorporated on next bulk build, thanks!

jrmarino:
push, thanks

PR #15, opened 2015-08-07
Fix for net-im/tox

mneumann:


jrmarino:
this is an unnecessary patch to "Makefile".
SODIUM_CONFIGURE_ON=	--with-libsodium-libs=${LOCALBASE}/lib \
			--with-libsodium-headers=${LOCALBASE}/include

can be added to Makefile.DragonFly thus avoiding a patch (which we want to avoid)

mneumann:
Ah, nice! Much better! Thanks!

PR #16, opened 2015-08-08
Patches for net-im/qTox

mneumann:


jrmarino:
Why replace the current ffmpeg with one that is deprecated and will be removed at some point?

What's wrong with current ffmpeg?

jrmarino:
qtox builds without any patches once it was unblocked by libfilteraudio.

mneumann:
@jrmarino are you sure it builds without patches? On my system it won't. The qtox code is using a newer version of libavutil  (part of ffmpeg) and a macro AV_IS_INPUT_DEVICE (http://ffmpeg.org/doxygen/trunk/log_8h.html) was not defined on "old" ffmpeg.

jrmarino:
I'm sure it builds on our official builder, muscles, without patches.
It's also tested on all FreeBSD systems -- it has to build there out of the box as well.

mneumann:
ok, then please ignore my patch!

jrmarino:
it built on master, but fails on Release 4.2:
http://pkgbox64.dragonflybsd.org/bulk/Release42-default/20150812_131440/logs/errors/qTox-0.150614_1.log

It's just missing a define or header that is fixed on master.

PR #19, opened 2015-08-30
Make lang/rust-nightly work on DragonFly

mneumann:
This will replace the port lang/rust-dragonfly.


jrmarino:
Especially with this pkg-plist patch, I think most, if not all of this, should be sent "upstream" to ports.  That means getting with the rust maintainer (possibly via PR) and make the changes there.  

Also, are you confident nightly will not break often?  If this requires a lot of continuous tweaking to keep support on DragonFly, then I might prefer rust-dragonfly for stability.

mneumann:
Yes, I will send patches upstream. Also, this will allow to use a Makefile.dragonfly instead of the Makefile.diff. 

Atm, there is no new updates to rust-nightly in FreeBSD ports as I think they are figuring out how to compile Rust with LLVM 3.7 (which is required to build a more recent version). The problem is that Rust tries to statically link to LLVM, but starting from LLVM 3.7 it is shared library only. I tried to link to the shared libraries (which involves some patching), but then the problem turns out to that it requires RTTI (or some other strange linkage problem), and LLVM 3.7 disables this. Conclusion: The only easy way to get more recent Rust versions is to NOT use the LLVM from ports. But lets see what FreeBSD is doing. 

I am working on continuous builds for DragonFly. Once this is done, I think the rust-nightly port won't break often. The only issue that would break atm is that I have to manually patch the snapshots.txt file. This won't be neccessary once I have a buildslave running that sends back these snapshot information to the Rust git repo.

But for now, better keep rust-dragonfly.

mneumann:
Upstream request sent: https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=202869

mneumann:
Closing in favor of https://github.com/DragonFlyBSD/DeltaPorts/issues/23.

PR #21, opened 2015-09-10
games/freeminer: initial port

zrj-rimwis:
Connecting to servers, walking, sounds putting bricks works.
Everything else needs further testing.


jrmarino:
Seems to me that most of those patches could be removed with "USES+= alias" ?

zrj-rimwis:
No, there are few places with cpu affinity handling in __FreeBSD__ paths.
Currently it doesn't use cpu pinning on DragonFly.

PR #22, opened 2015-09-11
sysutils/conky: initial port with intent to upstream it

zrj-rimwis:
Default .conkyrc config is working.
Currently:
  no disk I/O (devstat fun);
  running proc count stays at 0;
  six digit pid numbers get truncated to just 5 ones.

Port is very complicated due to FreeBSD patches that prevents applying few changes now.
Once those files/* get settled-in (or dropped) and if there would be interest from users
planing to submit dragonfly/* patches upstream for official support.

Would be very glad for patches and feature testing.
Happy ~/.conkyrc pimping!

p.s. port is explicitly made to be upstream-happy(note autogen.sh)

### DIFF
--- a/ports/sysutils/conky/Makefile.DragonFly
+++ b/ports/sysutils/conky/Makefile.DragonFly
@@ -1,0 +1,6 @@
+USES+=  autoreconf:build libtool
+
+CFLAGS+= -Dstatfs64=statfs -Wno-maybe-uninitialized
+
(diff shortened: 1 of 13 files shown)


jrmarino:
This fails, check muscles.  Probably needs gettext.

jrmarino:
sorry, conky is fine, it was freeminer that failed.

PR #25, opened 2015-09-14
games:/freeminer: fix libintl.so detection on DragonFly

zrj-rimwis:
DragonFly is a BSD.
Why NLS option is coming by default?


jrmarino:
NLS is always on by default, like DOCS and IPv6.

PR #28, opened 2015-09-15
graphics/cal3d: fix memcpy/memset decl issue

zrj-rimwis:
cal3d tests:
  PASS: converter/skeleton
  PASS: converter/mesh
  PASS: converter/material
  PASS: converter/animation
  ==================
  All 4 tests passed
  ==================

cal3d-devel: unchecked(tests/Makefile gen issue)

### DIFF
--- a/ports/graphics/cal3d-devel/dragonfly/patch-src_cal3d_platform.h
+++ b/ports/graphics/cal3d-devel/dragonfly/patch-src_cal3d_platform.h
@@ -1,0 +1,12 @@
+--- src/cal3d/platform.h.orig	2006-03-02 00:55:35.000000000 +0200
++++ src/cal3d/platform.h
+@@ -77,6 +77,9 @@ typedef int intptr_t;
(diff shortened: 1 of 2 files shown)


zrj-rimwis:
Next try:
graphics/py-soya3d (everything I tested works)

Builds but:
games/balazar (works without sound, but problems with OGG streams), py-soya3d is fine with audio
games/balazarbrothers (no sound, not a big issue)
games/el (works but cannot connect to server, too old version - 193, needed 194)
games/slune (wrongly packaged or don't get it how to invoke it, also gettext.localisation(..).intstal(1) problem)

jrmarino:
can you resubmit without the ifdef __DragonFly condition?  just include <string.h> unconditionally.  This is something that can be pushed upstream.

jrmarino:
don't bother, i'll just commit to ports.

jrmarino:
Fixed upstream, pull request obsolete.

PR #29, opened 2015-09-15
D1509 add munger

zrj-rimwis:
Few simple fixes.


zrj-rimwis:
Build order:
lang/munger

then:
editors/dkns (tested works)
www/ephemera
net/wackford-squeers

PR #30, opened 2015-09-15
games/scummvm: fix PREFIX issue

zrj-rimwis:
For alternative paths(unlike standard /usr/local) they should be fixed
in ports Makefile (by upstream) with REINPLACE_CMD like in other ports.


jrmarino:
i don't like this because it's a regression.  I'd rather update the patch if it's not complete.

zrj-rimwis:
Hmm need to recheck, but what is a purpose of ${PREFIX} here?
In this context it should point to system default installation directory to get includes.
${PREFIX} isn't even defined when building standalone.
If it is only a matter of patching port Makefile, I can do it too in diffs/Makefile

jrmarino:
"standalone" has nothing to do with this.  we are patching for ports.
Hardcoding /usr/local is wrong, period.  It has to be variable.  (yes, it's wrong for FreeBSD and should be fixed there)

zrj-rimwis:
OK, I stand corrected patch-configure is bad in dports compilation too.
./configure diff between previous and mine: http://sprunge.us/NEdW
Using -I${PREFIX}/include is wrong since during configure it expands to:
c++ -lpthread -L/lib ... -pedantic ... -I/include ... -ljpeg
Naturally this will fail.

I can add: -e "s@/usr/local@${LOCALBASE}@g" to ${REINPLACE_CMD}
This works in my setup.

jrmarino:
All that proves is that you are missing an LDFLAG setting (that doesn't show on FreeBSD because it searches /usr/local/lib by default).

jrmarino:
I fixed this properly using the designated configuration switches.

jrmarino:
also, don't avoid patches by rule.   I some cases I would have preferred patches to dfly-patch target.  With the latter, you lose all the context.

PR #31, opened 2015-09-16
D1509 add ucommon

zrj-rimwis:
Few simple compile fixes, mostly alias.


zrj-rimwis:
Build order:
devel/ucommon
devel/commoncpp

On PASS:
audio/ccaudio2
devel/ccrtp
devel/libzrtcpp

PR #32, opened 2015-09-16
D1509 fix ti group

zrj-rimwis:
Simple fix with regex, hard with patches.
Idea is to include dragonfly ARCH under \*-\*-\*bsd\* group.
This changes -D__LINUX__ --> -D__BSD__ (for now mostly was cosmetic).
Also prevents wrongly dumping x86_64 directly to CFLAGS (libticables2).


zrj-rimwis:
devel/libtifiles2 (rebuild)
comms/libticables2
comms/libticalcs2
comms/tilp2
emulators/tiemu3 (tested, works)

Also try build,
emulators/tilem

PR #34, opened 2015-09-16
sysutils/freeipmi: add alias

zrj-rimwis:
Sadly cannot test actual connection to bmc, network separation issue.
Still looks promising.


jrmarino:
used "alias" instead, but it fails to package with QA checks.

PR #41, opened 2015-09-18
games/quadra: fix alloca.h and use nullptr

zrj-rimwis:
No sound, "Audio write: Resource temporarily unavailable"
Other SDL based games have it too.


jrmarino:
I think you want to "USE_CXXSTD=c++11" here, rather than raw CXXFLAGS.

zrj-rimwis:
Good point, result is the same.

PR #46, opened 2015-09-18
x11-themes/qtcurve-qt45: fix std::isnan issue

zrj-rimwis:
Nothing better came in mind.

First rebuild:
x11-themes/qtcurve (to fetch Makefile.DragonFly)

Only then try:
x11-themes/qtcurve-qt4
x11-themes/qtcurve-qt5


zrj-rimwis:
Hopefully this will be enough.

PR #48, opened 2015-09-19
games/torcs: simple fix

zrj-rimwis:
Why would USES+= alias be by default in port Makefile???


jrmarino:
I'll remove USES=alias upstream, thus I'll remove the first line in MD here.

PR #51, opened 2015-09-19
lang/yap: fix gcc5 build

zrj-rimwis:
While there add few missing __DragonFly__ checks


jrmarino:
does not work.
http://muscles.dragonflybsd.org/bulk/bleeding-edge-potential/20150919_071043/logs/errors/yap-6.2.2_1.log

PR #53, opened 2015-09-19
D1509 endian h pack

zrj-rimwis:
Few simple fixes, most of them tested for correctness


jrmarino:
I'm going to move some of these alias fixes upstream.

PR #55, opened 2015-09-20
sysutils/scalpel: add alias

zrj-rimwis:
Too simple to fail


jrmarino:
pushed upstream instead.

PR #57, opened 2015-09-20
math/vtk5: add alias

zrj-rimwis:
Build:
math/vtk5
x11-toolkits/vtkfox


jrmarino:
I'll push USES=alias upstream.

PR #58, opened 2015-09-20
cad/meshdev: add cstring for memcpy

zrj-rimwis:


jrmarino:
pushed upstream instead.

PR #59, opened 2015-09-20
sysutils/usrinfo: add string.h

zrj-rimwis:


jrmarino:
pushed upstream instead.

PR #61, opened 2015-09-20
games/ppracer: TUX

zrj-rimwis:


jrmarino:
pushed upstream instead

PR #63, opened 2015-09-20
math/ised: add cstring for strlen

zrj-rimwis:


jrmarino:
pushed upstream instead

PR #64, opened 2015-09-20
graphics/graphot: add cstring

zrj-rimwis:
Tested


jrmarino:
pushed upstream instead

PR #65, opened 2015-09-20
converters/fix-mime-charset: use cstring and cstdlib

zrj-rimwis:


jrmarino:
pushed upstream instead

PR #66, opened 2015-09-20
graphics/ampasCTL: use ctdlib for alloca

zrj-rimwis:


jrmarino:
pushed upstream instead

PR #68, opened 2015-09-20
games/galaxyhack: add missing cstring for std::memcpy

zrj-rimwis:


jrmarino:
pushed upstream instead.

PR #69, opened 2015-09-20
games/moria: add alias

zrj-rimwis:
It works!!!


jrmarino:
pushed upstream instead.

PR #70, opened 2015-09-20
games/crack-attack: add cstring for strncpy

zrj-rimwis:


jrmarino:
pushed upstream instead.

PR #71, opened 2015-09-21
devel/clanlib1: add alias

zrj-rimwis:
Then try build:
games/openalchemist
games/trophy


zrj-rimwis:
Identical USES+= alias allows to build devel/clanlib too,
but since it's only usage case games/super_methane_brothers
fails to start due to missing protocol in dfly I see little point to build it.

jrmarino:
pushed upstream instead.

PR #75, opened 2015-09-22
lang/gauce: small fix

zrj-rimwis:
make regression-tests:
Total: 14499 tests, 14498 passed,     1 failed,     1 aborted.

Single gauche--net.so test fails.


zrj-rimwis:
Build:
lang/gauche

Then:
devel/gauche-readline
graphics/gauche-gl        (OpenGL 3D examples work)

PR #76, opened 2015-09-22
science/cdcl: expand OS detection

zrj-rimwis:
Graphical demos work


zrj-rimwis:
Then:
science/rubygem-ruby-dcl

PR #79, opened 2015-09-22
japanese/Wnn6-lib: add .so.2.0 to list too

zrj-rimwis:
Avoids installing libs with broken symlinks.
Also satisfies LIB_DEPEND for japanese/esecanna-module-wnn6


zrj-rimwis:
Build:
japanese/Wnn6-lib (rebuild)

Then:
japanese/esecanna-module-wnn6

PR #86, opened 2015-09-22
games/frozen-bubble: simple fix

zrj-rimwis:
Drop -Qunused-arguments flag on gcc


jrmarino:
generic fix pushed upstream

PR #88, opened 2015-09-30
net/openslp: add alias

zrj-rimwis:
Needed for sockets

gmake check:
  PASS: Success!
  PASS: slp-conf-test
  PASS: slp-compare-test
  ==================
  All 2 tests passed
  ==================


jrmarino:
pushed upstream instead.

PR #90, opened 2015-09-30
x11-themes/qtcurve: prune isnan patch

zrj-rimwis:
Untested.
Seems no longer needed, should unbreak qtcurve-gtk2/qtcurve-kde4 builds.


jrmarino:
do we need to revert this prune?   qtcurve slaves are failing because of isnan again, on pkgbox64 as well.

zrj-rimwis:
it is a \#include cmath vs math.h issue,
trying to cook something better, building qt deps takes many SBUs :)

PR #91, opened 2015-10-01
net-mgmt/wifimgr: install fix

zrj-rimwis:
Workaround against addition of MAKE_ENV= INSTALL="intstall -U"


jrmarino:
variation of fix is now upstream

PR #95, opened 2015-10-01
devel/libdisorder: mark as jobs_unsafe

zrj-rimwis:


jrmarino:
pushed upstream instead.

PR #98, opened 2015-10-01
lang/beignet: fix build

zrj-rimwis:


zrj-rimwis:
yup, will redo later when dports get updated

PR #100, opened 2015-10-02
devel/gdb: just to build

zrj-rimwis:
Once again I thrive on diffs.
Not a proper but allows to step through code.
"layout split" still able to display source too.
So if I didn't missed it should build and allow someone to finish the job.


zrj-rimwis:
When stepping by asm instruction it gets lost(loops) in ld-elf.so.2.
Few ?? in libc.so but those are in base gdb as well.

PR #108, opened 2015-10-03
sysutils/deltup: simple fix

zrj-rimwis:
No tests


zrj-rimwis:
also try:
sysutils/getdelta

PR #111, opened 2015-10-03
x11/cool-retro-term: add OPSYS

zrj-rimwis:
UNTESTED_REASON= QT5(requires dbus)


jrmarino:
Ugg, this port is ugly.  And it's probably going to change since FreeBSD 8 is no longer supported.
I wonder why CC=clang is explicitly defined.  That's strange.  maybe some code checks value of "CC" but that's pretty bad.

jrmarino:
i guess danfe isn't going to change this any time soon so I'll take this patch.

PR #114, opened 2015-10-04
www/quickie: fix build

zrj-rimwis:
Add workaroung for seolim -I


zrj-rimwis:
Not pretty but OPTIONS_UNSET= DOCS is a no-no?

PR #116, opened 2015-10-04
graphics/zphoto: add alias

zrj-rimwis:


jrmarino:
fixed pushed upstream

PR #118, opened 2015-10-04
archivers/hpack: add alias

zrj-rimwis:
It all right I'm from EU


jrmarino:
fixed pushed upstream.

PR #123, opened 2015-10-10
Add initial wayland and weston ports for DragonFly

ivadasz:
An initial port of libwayland and the weston reference compositor to DragonFly.
Currently DragonFly specific because of, e.g. usage of libdevattr for device enumeration.

for testing with x11 backend, run something like:
    mkdir /tmp/wayland_xdg
    chmod 0700 /tmp/wayland_xdg
    env XDG_RUNTIME_DIR=/tmp/wayland_xdg weston --use-pixman --backend=x11-backend.so

for testing with drm backend, no Xorg server may run, and only a single instance of weston with drm-backend can run at a time (This also requires that the i915kms or radeonkms driver is running for graphics):
- Create a group weston-launch
- Add your user to this group weston-launch
- Then run:
    mkdir /tmp/wayland_xdg
    chmod 0700 /tmp/wayland_xdg
    env XDG_RUNTIME_DIR=/tmp/wayland_xdg weston-launch -- --use-pixman --backend=drm-backend.so

important TODOs left:
  - Automatically create a weston-launch group (for selecting users which are allowed to use
    weston-launch to run the weston compositor via the setuid weston-launch binary)


M4SSD35TRUCT10N:
@ivadasz Excellent! I wished that day to come and now it is true. Great work!

PR #126, opened 2015-10-11
Fix audio/audacity

mneumann:


jrmarino:
this can't be correct:

++#if NEED_ULONG || __FreeBSD__ || __DragonFly__


__DragonFly__ has no value   (unlike __FreeBSD__)

You need to check if these are defined, not their value

mneumann:
Then 
```
|| defined(__DragonFly__)
```

should work. Actually, it worked without, probably because ```NEED_ULONG``` is set.

PR #127, opened 2015-10-11
Fix x11-toolkits/swt-devel

mneumann:


jrmarino:
why was Makefile.diff removed?

There is no native untar, removing this diff should cause the port to fail.  (Why doesn't it?)

jrmarino:
ah, you moved it to Makefile.DragonFly

jrmarino:
I believe the second change is wrong.
You are looking at the ports version. the "amd64" callouts are transformed to x86_64 during merging for the entire dports tree.  adjusting for adm64 should not be necessary.

jrmarino:
I'll remove this from M.D.

PR #128, opened 2015-10-13
F13 xorg sync

zrj-rimwis:
Unless something new will be introduced in mainline catch up
these should apply cleanly.


jrmarino:
Thanks.  All we had to do was wait one week. :)

zrj-rimwis:
I'm almost tempted to know why it wasn't done before given that it certainly wasn't a libjpeg-turbo issue.
Maybe smth to do with cygwin bump to xorg 1.17.2 on 12th.
But all that matters is that it got updated and we got very similar thing they merged to :)
Also going to drop Makefile.diff in net/tigervnc (no longer needed).

So now only xorg-drivers will need adjustment (I'd prefer to have ours current minimal set).
Another issue will be graphics/dri DRI/VDPAU handling move to graphics/libGL/Makefile.common (minor mesa bump).
It complicates lib/vdpau/*.so installation.
From their current variant looks like these are intended for new port (maybe graphics/mesa-vdpau ?).
But since they lack drm support and just ".if 0" it I think it is sensible to continue providing vdpau drivers in graphics/dri (already in pkg-plist).

jrmarino:
hi zrj,
Muscles *just* went down, but before it I, I noticed graphics/dri didn't pass configure because vdpau wasn't found.  When I pre-built libGL last night, I suspected that the default to build vdpau (which is unique to dragonfly) was lost, but I wasn't sure because frankly there seems to be old and new versions of xorg framework embedded.

Can you check graphics/libGL and verify vdpau is not longer building by default and provide a fix?  You may have to wait for swildner to recover muscles to see the log, but you can probably verify this yourself with the latest versions of deltaports and freebsd ports.

jrmarino:
P.S. tigervnc was fixed in ports last night

zrj-rimwis:
Hi marino,
dri/vdpau issue came because of http://svnweb.freebsd.org/ports?view=revision&revision=399021
Not sure how to fix it properly :( can't see what Delta scripts did after import, so it is hard to guess.
Maybe for now just disable vdpau option like free do?
All that would be lost is vdpau drivers for radeons. libvdpau would be unaffected, just not functional without drivers.
I was monitoring build as it went down. graphics/dri didn't fetch dependency on libvdpau.so from new graphics/libGL/Makefile.common. 
OK I'll ask swildner and try few builds with Makefile.common changes under COMPONENT:Mdri.

Btw have regen patch-configure for multimedia/xvid, but imho sed -e "s/\*bsd\*|/*bsd*|dragonfly*/g" configure would be more universal.

jrmarino:
we need to keep vdpau; ftigeot added it.  I fixed it on your newports but that fix was deleted with the rest.

I don't think it's that hard to fix (one or two lines) but I am tired of chasing the breakage and figured you could find it quickly too.

zrj-rimwis:
ok i'll try, cleaning my radeon box first.
btw muscles came back online and build still continues.

PR #131, opened 2015-10-15
mesa/vdpau: package radeon vdpau libs in dri

zrj-rimwis:
Missed this hunk, vdpau libs were compiled but
VDPAU="@comment" took precedence against VDPAU="" in PLIST_SUB
thus radeon vdpau libs were not packaged.

### DIFF
--- a/ports/graphics/libGL/diffs/Makefile.common.diff
+++ b/ports/graphics/libGL/diffs/Makefile.common.diff
@@ -8,7 +8,13 @@
  WRKSRC=			${WRKDIR}/mesa-${MESADISTVERSION}
  DESCR=			${.CURDIR}/pkg-descr
  PLIST=			${.CURDIR}/pkg-plist
-@@ -134,11 +135,12 @@ CONFIGURE_ARGS+=--disable-vdpau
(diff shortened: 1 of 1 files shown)


zrj-rimwis:
Also having strange issues with make in staged branch. (master branch is fine)
When building dep ports it aborts with "make -j requires positive integer" (MAKE_JOBS_UNSAFE=yes helps)
When building actual port [graphics/dri] i get insane load: 120 60 40 (still builds because i have 16gb of ram)

zrj-rimwis:
ah it should still be _SMP_CPUS!= instead of _MAKE_JOBS_NUMBER!= ${SYSCTL} ...
https://github.com/DragonFlyBSD/DeltaPorts/blob/master/special/Mk/diffs/bsd.port.mk.diff#L157
Current variant has empty _SMP_CPUS and passes too all makefiles just plain "-j"

jrmarino:
okay, that's a manual merge cut-n-paste error.  I can't easily connect to server where I am.  If you want, you can submit a PR to fix the diff in DeltaPorts but I'll have to fix DPorts manually.

Otherwise do nothing and I'll fix it when I can (maybe tonight)

PR #132, opened 2015-10-15
x11-themes/qtcurve: deep patch

zrj-rimwis:
This seems to be a proper one, all building combos pass.
If it fails again I give up.


zrj-rimwis:
Another dumb meta port.
Lets try testing it fully now:
First try build: x11-themes/qtcurve (should build all 4 flavours + utils)
Then retry building all x11-themes/qtcurve-* separately.
And just to be sure again retry x11-themes/qtcurve (should not rebuild anything)

PR #133, opened 2015-10-16
x11-drivers/xf86-input-vmmouse: fix ioperm detect

zrj-rimwis:
Just simple configure patch to use /dev/io
Correct?


jrmarino:
looks reasonable

PR #136, opened 2015-10-16
devel/libosmo-sccp: fix libosmo endian.h

zrj-rimwis:
Since osmocom/core/endian.h is installed into local
USES+= alias doesn't help and actual patch is needed.


zrj-rimwis:
What is a point of including USES+= alias to freebsd ports?
I would prefer to see it in Makefile.Dragonfly rather then in buildlogs
(specially on cmake ones that hide command lines)
devel/libosmocore should be patched directly.

Build:
devel/libosmocore (rebuild), then all dependent on it.

PR #137, opened 2015-10-16
comms/nmealib: mark jobs unsafe

zrj-rimwis:
fails even with -j2
*sometimes


jrmarino:
pushed upstream (all not-safe is pushed upstream once it's demonstrated to be repeatable)

PR #141, opened 2015-10-17
Build wayland-egl support into libEGL

ivadasz:
gl-renderer stuff can best be tested, by running the nested commpositor. That will use the llvmpipe software gl-renderer from mesa, which is fine for testing. Trying to use hardware acceleration just gives a stuck render ring (on i915kms); or a somewhat working desktop with lots of corruption which causes a hard freeze after a couple of seconds (on radeonkms).


ivadasz:
I'm not really sure if this usage of port options makes much sense.
Also, this needs a review for which mesa ports this WAYLAND option is actually relevant ...

jrmarino:
should we close this request?  or what's going on with it?

PR #143, opened 2015-10-31
Fix aircrack-ng

shamazmazum:
Hi. This is related to https://github.com/DragonFlyBSD/DPorts/issues/164

Please read that for explanation

### DIFF
--- a/ports/net-mgmt/aircrack-ng/Makefile.DragonFly
+++ b/ports/net-mgmt/aircrack-ng/Makefile.DragonFly
@@ -1,0 +1,9 @@
+USES += alias
+
+dfly-patch:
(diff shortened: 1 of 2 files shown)


jrmarino:
it doesn't make sense to both have Makefile.DragonFly and patch Makefile.  It should be one or the other.

You can move the diff to Makefile to Makefile.DragonFly by adding to the target "dfly-patch:"

shamazmazum:
Fixed Makefile.DragonFly. Also a new patch to source I wanted to add is now in "dragonfly" directory, not in "diffs".

jrmarino:
looks better, I'll try it after current bulk build finishes.

PR #148, opened 2015-11-03
devel/libdap: remove crippling

zrj-rimwis:
Disable EXTRA_PATCHES+=extra-patch-dds.yy

gmake check:
All 40 tests were successful.
All 29 tests were successful.
All 118 tests were successful.
All 170 tests behaved as expected.
  1: getdap -d http://test.opendap.org/dap/data/nc/fnoc1.nc getdap-testsuite/fnoc1.nc.dds (pass) ok


jrmarino:
upstream error corrected instead.

PR #152, opened 2015-11-05
Fix asterisk-11.20.0 compilation

dkgroot:
Based on FreePBX Ports Collection

### DIFF
--- a/ports/net/asterisk11/Makefile.DragonFly
+++ b/ports/net/asterisk11/Makefile.DragonFly
@@ -1,0 +1,19 @@
+OPTIONS_DEFAULT=BACKTRACE CURL BASE GSM LUA MP3PLAYER NEWT PGSQL SNMP SPEEX SQLITE2 UUID VORBIS
+#OPTIONS_SINGLE_COMPILER:= BASE 
+
+dfly-patch:
+	${REINPLACE_CMD} \
+		-e 's@\(#if !defined(__FreeBSD__)\)@\1 \&\& !defined(__DragonFly__)@g'		\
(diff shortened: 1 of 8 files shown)


jrmarino:
okay.  There was no reason to change STATUS but the impact is minimal so I'll take the whole request, thanks.

jrmarino:
PORTS_OPTIONS is not getting redefined so it's not building.  It probably requires :N modification (which I will attempt)

jrmarino:
i think the use of <bsd.port.pre.mk> is breaking this, it's going to require a makefile diff

dkgroot:
It seems to work fine inside dports. I only wanted to change the default value the compiler being used (to base instead of ports/gcc. Should i make a change, or has this already gone through ?

jrmarino:
I suspect you saved your options.  DADHI was forced on no matter what.  It's only possible not to have it if options set manually.

dkgroot:
BTW: Changing the defaults was not really required. So if you like, we can remove the line completely (maybe better)

jrmarino:
it is required.  asterisk11 cannot have dahdi as an option.

I did assume you had a reason for the change

dkgroot:
I wanted to change the default compiler (which get's set to ports/gcc instead of base/gcc by the original Makefile). It does work with either though. So If it's better to remove the line completely i am fine by that. Sorry about this.

jrmarino:
Ok, it's pushed.
btw, there's not difference between GCC and BASE for dragonfly so we could have left it as gcc.

dkgroot:
Ah ok. Thanks for the info ! Same will apply to the asterisk13 commit. Would you like me to update it ?

dkgroot:
Thanks again for you help!

PR #162, opened 2015-11-23
Fix build of sysutils/multitail on DragonFly.

ddbilik:


jrmarino:
it is in ports now; it was fixed upstream: https://github.com/DragonFlyBSD/DPorts/commit/cf6e651c5b79d66a6851d05342b1abf7f07872ea

PR #180, opened 2015-12-02
print/ghostscript8-base: add DragonFly under BSD

zrj-rimwis:


jrmarino:
i hate that there are at least 4 versions of ghostscript in ports.  Do we need them all?  (bapt says each one addresses a certain need particularly with CJK but it sounds like a crapfest to me)

PR #182, opened 2015-12-02
Fix build of ftp/tnftpd on DragonFly.

ddbilik:


jrmarino:
FYI, the base "ftp" is the same as this port (from upcoming release 4.4 and master)

ddbilik:
Oh, I should have checked before producing a duplicate "product". Next time. :) Thanks for info.

jrmarino:
it's not duplicate per se.  At the current time, our FTP is the same as the port, but the port could easily be newer since base tool versions are locked in and normally don't change for month or even years.

PR #183, opened 2015-12-03
Fix build of x11-fm/qtfm on DragonFly.

ddbilik:
DragonFly is missing EVFILT_FS implementation in kqueue(2). So rather than
checking for mount events inefficiently (for example via getmntinfo(3)),
disable autoBookmarkMounts() feature for now (until there is EVFILT_FS).

### DIFF
--- a/ports/x11-fm/qtfm/Makefile.DragonFly
+++ b/ports/x11-fm/qtfm/Makefile.DragonFly
@@ -1,0 +1,1 @@
+USES += alias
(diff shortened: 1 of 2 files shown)


ddbilik:
> so we definitely shouldn't be calling kevent in that case

I've tested this and it behaves correctly, but you're right... Initializing ki with bzero() will make kevent() calls predictable, while keeping the desired functionality (ie. "auto-enable" that code when EVFILT_FS is available). Should be fixed with pull request #185.

Thanks for spotting this.

PR #191, opened 2015-12-04
Fix build of textproc/bsdgrep on DragonFly.

ddbilik:
Currently, there is no libgnuregex equivalent on DragonFly, so remove
GNU_COMPAT from OPTIONS and explicitely set WITHOUT_GNU_COMPAT.


ddbilik:
Thanks. I also see OPTIONS_EXCLUDE used in several places... Does it have the same effect? And which approach is more "proper"?

jrmarino:
no, OPTIONS_EXCLUDE doesn't really work well.  There is often logic that works based on the state of an option (on or off) but EXCLUDING removes the option completely, often breaking the code.  There needs to be a new type of exclusion but it doesn't exist yet.

ddbilik:
Thanks for clarifying this. Fixed in pull request #192.

PR #196, opened 2015-12-04
games/ufoai: Fix for data extract

zrj-rimwis:
Data is bigish ~1GB
Game runs, playing is fine on i915 haswell

### DIFF
--- a/ports/games/ufoai-data/Makefile.DragonFly
+++ b/ports/games/ufoai-data/Makefile.DragonFly
@@ -1,8 +1,7 @@
 
-# attempt to fix extract
-WRKSRC?= ${WRKDIR}
+# zrj extraction issue
+.undef NO_WRKSUBDIR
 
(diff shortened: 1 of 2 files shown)


jrmarino:
NO_WRKSUBDIR is set in the makefile so this line has no effect.

This port was building fine before, it shouldn't need anything other than removing the ignore.

zrj-rimwis:
huh, then it gets somehow unset?

jrmarino:
no.  It should be working fine.  Are you using a clean poudriere build?  WRKDIR stuff changed a few weeks ago.  You have to have the last tree.

jrmarino:
if you remove the WRKDIR redefinition from MD, I'll merge it in.  I'm confident it it work.

jrmarino:
fixed a different way (a way that should be upstreamed)

PR #197, opened 2015-12-04
games/vegastrike: Fix build, unignore data

zrj-rimwis:
Also fixed main Makefile

Game works, flying, laser shooting, sounds OK


jrmarino:
I pushed the makefile fix upstream so I'll have to delete that one.

PR #220, opened 2015-12-05
games/quake2-ctf: fix extract

zrj-rimwis:
This is just capture-the-flag mod part for quake2


jrmarino:
there's something amiss here, I'll take a look.  unzip and -xf don't go together.

jrmarino:
fixed a different way.

PR #230, opened 2015-12-06
games/sauerbraten: fix build

zrj-rimwis:
Cool CUBE2 based FPS shooter game works great on haswell
Note: quite big 655M installed

### DIFF
--- a/ports/games/sauerbraten/Makefile.DragonFly
+++ b/ports/games/sauerbraten/Makefile.DragonFly
@@ -1,0 +1,6 @@
+
+MANUAL_PACKAGE_BUILD=	huge (655 MiB)
+
+.if defined(DPORTS_BUILDER)
+.undef MANUAL_PACKAGE_BUILD
+.endif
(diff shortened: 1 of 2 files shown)


jrmarino:
655M is too big, this needs MANUAL_PACKAGE_BUILD set with the typical DPORTS BUILDER exception

zrj-rimwis:
Honestly don't know how to deal with this one.
Single distfile contains both engine and data (563M).
Package .txz is 540M, installed 655M
Maybe we need some marker to build on muscles but ignore on pkgbox?
Or maybe even smth like PASSTHROUGH=yes that ignores build but includes in dports as is.
Would be usefeul for ports like p5-OpenGL too

jrmarino:
"Maybe we need some marker to build on muscles but ignore on pkgbox?"

That's the the dports builder exception does.  Look at any of the huge -data ports.  You many have to set MANUAL_PACKAGE_BUILD in Makefile.DragonFly first.   (or set it if not DPORTS_BUILDER)

jrmarino:
I'll add the filter.

jrmarino:
ah, i missed that you already did!

PR #234, opened 2015-12-06
games/ezquake: fix build

zrj-rimwis:
Ugly port, low-res, needs original game data


jrmarino:
why not just IGNORE it then?

zrj-rimwis:
What's the difference between broken and ignore?
i thought ignore just skips build while broken indicated that it is broken at runtime

zrj-rimwis:
updated with IGNORE :)

PR #252, opened 2015-12-07
Add DragonFly support to sysutils/monit.

ddbilik:
Take these FreeBSD support files as a reference...

libmonit/src/system/os/freebsd/Link.inc
src/device/sysdep_FREEBSD.c
src/process/sysdep_FREEBSD.c

... copy them to DragonFly counterparts and adapt initprocesstree_sysdep()
and used_system_memory_sysdep() in the latter.

Note: optimally, we should get this included in Monit source, however DPorts
is an ideal place to test and polish it before pushing to upstream.

### DIFF
--- a/ports/sysutils/monit/Makefile.DragonFly
+++ b/ports/sysutils/monit/Makefile.DragonFly
@@ -1,0 +1,1 @@
+USES += autoreconf libtool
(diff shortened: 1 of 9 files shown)


ddbilik:
Though it was merged, STATUS hasn't changed and the port has not get to DPorts... Does it have any build problem?

jrmarino:
no, it's fine.  I just missed it.  It's not really a big deal because the next bulk build would have caught it anyway.  It is in DPorts now.

PR #255, opened 2015-12-07
games/nexuiz: fix port

zrj-rimwis:
Darkplaces mod plays awesomelly! Both glx and sdl backends
on both radeonsi and haswell.

WARNING: size..
distfiles - single engine+data(888M) + optional(ON) map pack(108M)
package - .txz 879M
installed - 963M


jrmarino:
too big, it needs the dports builder limitation

jrmarino:
the MANUAL_PACKAGE_BUILD is set in ports, so I'll add a dports builder exception.

PR #287, opened 2015-12-10
Fix build of net/wmping on DragonFly.

ddbilik:


ddbilik:
Thanks for the hint, I've just checked that it's sufficient to patch configure directly. I'll prepare it into another pull request.

PR #307, opened 2015-12-13
D1512 prep mesa1107

zrj-rimwis:
For upcoming mesa 11 bump


jrmarino:
so if I read this correctly, we can now bring the current versions in ports in dports without having to resort to LOCKing.  Is that correct?

zrj-rimwis:
Basically only thing that i found out was clover on radeons.
That extra patch would help for now to run OpenCL apps
like fft2d, clinfo for libMesaOpenCL.so as root.
Still thinking how to properly add /dev/dri/renderD device nodes in drm.

Also if mesa-11.0.7 build will be successful, after testing planning to add wayland-egl bits that ivadasz tried to PR.
Using same method as for libelf deps in libGL/Makefile.common.
But only for DFly_ver >= 4.5.
First need to explore how to do micro bumps for libdevq

PR #322, opened 2015-12-14
net/asterisk-g72x: add dragonfly

dkgroot:
from ports collection


dkgroot:
All the information is in the ports Makefile, what should i be pushing to just take everything from FreeBSD ?

jrmarino:
I don't understand what you are doing.
The commit message says "add dragonfly".
The only file added is "Makefile.DragonFly".
It is listed a binary and shows up blank on github.

1) it should not be binary
2) it should not be blank

So I am very much confused.  Maybe you having git issues?

dkgroot:
Ok That's not good.
If i just need something ported from the Portscollection, without any requirements on the dragonfly side, what should i check in ? Just the directory ?
What would you like the commit line to say ? I guess i don't know how to make just the port happen.

jrmarino:
"what should i check in ? Just the directory ?"

Nothing.

Everything that builds is brought into dports automatically with no special instructions.
In other words, if it's not in, it's either:
1) brand new
2) it fails to build
3) one of it's dependencies fails to build so the build is never attempted.

dkgroot:
Ok i did not know that. Will drop both Pull Requests. Sorry about this.
net/asterisk-g72x does build when running it manually in dports BTW. Is there anyway to see why it was not taken ?

jrmarino:
always check: 
http://muscles.dragonflybsd.org/latest-failures/

however, it's not listed.
so I check dports:
oh look, there it is:
http://gitweb.dragonflybsd.org/dports.git/tree/HEAD:/net/asterisk-g72x

Try pulling latest dports.

dkgroot:
Thanks :-)

PR #323, opened 2015-12-14
net/asterisk-chan_sccp: Add Dragonfly

dkgroot:
from ports-collection


jrmarino:
also appears corrupt

jrmarino:
this port is only 1 day old.
It would be imported on the next full bulk build of dports (along with all the other brand new ports that build on dragonfly)
However, I just did it manually.

dkgroot:
I know, i wrote it :-) Took a while for it to be commited to dports. Thanks for pulling it manually ! Sorry for the confusion !

PR #329, opened 2015-12-17
graphics/blender: update to 2.76b version

zrj-rimwis:
BMW v27 performance test is much better.
Even game engine is almost usable.


zrj-rimwis:
Ah we misunderstood each other.
I meant to bump version while fixing plist to correctly reflect port options.
There will be users who will run it only in HEADLESS mode.
Now it should be build fine.

PR #332, opened 2015-12-18
multimedia/avidemux: retweek ffmpeg inlining in cxx

zrj-rimwis:
This one is complicated too.

Rebuild just multimedia/avidemux first.
Then in separate run rest avidemux-{cli,plugins,qt4}

Tested, plugins now work.


jrmarino:
alright, but by definition all the patches in dports are dragonfly - specific.

zrj-rimwis:
Yeah, but in case someone comes across this on google, don't want them to have funny ideas ;)

PR #349, opened 2015-12-21
devel/fpc-bfd: redirect dep to binutils

zrj-rimwis:
devel/libbfd is build from an old binutils tarball.
It does have dragonfly target in config.bfd.
Also conflicts if any ports depend on devel/binutils (libbfd.a).

Thus on DragonFly it useful to avoid port conflicts.


jrmarino:
I've fixed this at freebsd level already.

PR #350, opened 2015-12-21
emulators/qemu-devel: Add vknet network backend for DragonFly's vknetd(1)

ivadasz:
Adds a vknet network backend to qemu-devel that allows utilizing the vknetd(1) userspace switch similarly to how it can already be used vor DragonFly vkernels.

can be used by adding to the qemu command-line something like:

-netdev vknet,id=vke0 -device virtio-net,netdev=vke0

which would add a virtio network device for the guest, which is connected to the outside via the default vknetd socket ( /var/run/vknet ). The vknetd socket can also be explicitly specified, e.g. with:

-netdev vknet,id=vke0,sock=/var/run/myvknet -device virtio-net,netdev=vke0

### DIFF
--- a/ports/emulators/qemu-devel/diffs/Makefile.diff
+++ b/ports/emulators/qemu-devel/diffs/Makefile.diff
@@ -1,0 +1,47 @@
+--- Makefile.orig
++++ Makefile
+@@ -24,7 +24,7 @@ ONLY_FOR_ARCHS=	amd64 i386 powerpc powerpc64 # XXX someone wants to debug sparc6
+ 
+ OPTIONS_DEFINE=	SAMBA X11 GTK2 OPENGL GNUTLS SASL JPEG PNG CURL \
+ 		CDROM_DMA PCAP USBREDIR GNS3 X86_TARGETS \
+-		STATIC_LINK DOCS
++		STATIC_LINK DOCS VKNET
+ SAMBA_DESC=		samba dependency (for -smb)
+ GNUTLS_DESC=		gnutls dependency (vnc encryption)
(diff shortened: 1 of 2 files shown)


jrmarino:
all the makefile.diff stuff has to be moved to Makefile.DragonFly

fleshenough:
Cool. :)

jrmarino:
The vknet patch can't apply at all (not even close) on the latest qemu-devel:
http://muscles.dragonflybsd.org/bulk/bleeding-edge-potential/20160111_025118/logs/errors/qemu-devel-2.5.0.log

If you can't supply a new one, the choices are:
1) remove the patch
2) freeze qemu-devel at the current version forever

ivadasz:
The update is definitely trivial for vknet-patch. Apparently nothing much has changed in the relevant qemu upstream code, so the patch should only need adjustments for line numbers, and context changes.
I'll probably find time for doing this (and verifying that the vknet backend still works), during the weekend.

You could of course disable the VKNET option by default until the patch is fixed, to do the 2.5 update immediately.

That failure on muscles.dragonflybsd.org looks like it's not stripping away the a/ and b/ prefixes in the file paths in the vknet-patch file.

jrmarino:
I regenerated the path and manually removed a/ and b/
I made it a regular patch that could be applied with EXTRA_PATCHES.  
As it was originally provided was unusable for me.  When you provide a new version, make sure it is in the same format as it is correctly in DeltaPorts/DPorts.

jrmarino:
It's best if you generate the patch with either ports-mgmt/genpatch tool or "make makeplist" and then change the filename to match what it is now.

PR #352, opened 2015-12-21
editors/lazarus: get it to build

zrj-rimwis:
Now ide builds, installs and runs.
There are quite many units that are ignored for dragonfly,
but since pkg-plist is dinamic, hard to say what needs to patched too.

Tested building from dports on live system, lets first see if muscles will like it.


zrj-rimwis:
Also fpc-regexpr might be missing Package.fpc.
Had to manually patch two makefiles, because fpcmake couldn't find Package/Makefile.fpc

PR #359, opened 2015-12-30
biology/htslib: fix build after libcurl addition

zrj-rimwis:


jrmarino:
already fixed upstream.  you should ask to save yourself wasted time.

PR #361, opened 2015-12-30
devel/love: fix build after version bump

zrj-rimwis:
LUA-JIT dependency needs recheck by someone w/ lua knowledge.
GUI starts and works, but prone to segfault in lua.so upon exit.


jrmarino:
isn't this already fixed upstream?

jrmarino:
apparently not.

PR #384, opened 2016-01-13
math/octave: regen patches on new version

zrj-rimwis:
GUI starts, seems functional.
Drop MAKE_JOBS_UNSAFE, tested with:
MAKE_JOBS_NUMBER={1,2,4,7,17,32}

There are 97 skipped ports.


jrmarino:
octave is a huge mess, freebsd is scrambling to fix.  The new 4.0 was pushed without testing any of the 103 dependencies.  It broke *ALL* of them, and now maybe it's fixed but many dependency ports are being marked broken.

zrj-rimwis:
well this PR only regens the previous patches and allows it to compile
So GNU_HOST=${CONFIGURE_TARGET} is more preferable than
${ARCH}-portbld-${OPSYS:tl} ?

Now moving to tbb and broken games

jrmarino:
iverilog is a makes jobs issues and it's been fixed upstream already.
Everything else is untouched, so go for it.

PR #388, opened 2016-01-13
science/gnudatalanguage: regen patches

zrj-rimwis:
regression-test still pass


jrmarino:
why are ".bak" files being patched?

Normally those aren't compiled at all.

zrj-rimwis:
ah i thought i renamed them to file.ext.intermediate
in this case it doesn't change two previous patches

PR #398, opened 2016-01-19
Mk/bsd.commands.mk: redef AWK in case of short form

zrj-rimwis:
This fixes shebangs at least in devel/libcheck:bin/checkmk
AWK?=awk comes from share/mk/sys.mk


jrmarino:
why this?  ".if defined(AWK)"

There's no possible way that ever evaluates to false, right?

jrmarino:
in other words, ".if  ${AWK:Nawk}" is enough I think

zrj-rimwis:
yeah, ".if ${AWK:Mawk}" is better

PR #413, opened 2016-01-26
sysutils/hal: to prevent gnome fallout after isofs/cd9660 changes

zrj-rimwis:
After leaf/~zrj/cd9660_changes, for now just on pending state
cause i need to acquire base version bump token.

Basically currently hal compilation relies on __FreeBSD_version
evaluating as 0, that is bad.

Btw synth always recompile devel/nspr and friends ..


jrmarino:
until the libraries are privatized, synth is going to rebuild some ports unnecessarily.  Basically the problem has always been there but synth is highlighting it.

PR #414, opened 2016-01-31
No need for libdevq in loader_get_pci_id_from_fd(), but use it in loader_get_device_name_for_fd()

ivadasz:
The "graphics/libGL: Implement loader_get_device_name_for_fd() via libdevq" change could probably be upstreamed to fbsd ports. For the other change upstreaming to ports likely depends on whether the HAVE_LIBDRM code for loader_get_pci_id_from_fd() works on older still supported versions of freebsd.


ivadasz:
Or rather just use https://patchwork.freedesktop.org/patch/68748/ instead of my "graphics/libGL: Implement loader_get_device_name_for_fd() via libdevq" change. That way libdevq could be removed as a dependency from libGL.

ivadasz:
The "graphics/libGL: Remove HAVE_LIBDEVQ code for loader_get_pci_id_for_fd()." change doesn't work correctly in all cases

PR #415, opened 2016-02-06
Fix build of devel/readline after privatization of ncurses.

ddbilik:
Explicitely remove -ltermcap from linker flags to ensure the port is buildable
even with TERMCAP option set.

### DIFF
--- a/ports/devel/readline/Makefile.DragonFly
+++ b/ports/devel/readline/Makefile.DragonFly
@@ -1,5 +1,6 @@
 USES+=	ncurses
-OPTIONS_DEFAULT:= ${DEFAULT_OPTIONS:NTERMCAP}
+OPTIONS_DEFAULT:= ${OPTIONS_DEFAULT:NTERMCAP}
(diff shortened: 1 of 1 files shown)


ddbilik:
Well, revising this once more, I think it's misleading to even present TERMCAP in options, when the system has no libtermcap to provide. I'll prepare another PR.

jrmarino:
is it blocking something?  we have a list of 100 ports that still need to be fixed (so it isn't news).  But is there a reason to bump this to the front of the line?

jrmarino:
maybe you don't know, but we are pushing our fixes to freebsd.
Although we have to be careful with something like readline.

ddbilik:
> is it blocking something?
> is there a reason to bump this to the front of the line?

Not really, I've just hit this when rebuilding ports after system upgrade. where I had TERMCAP option set from previous build.

I'm closing this, PR #416 is IMHO cleaner way to resolve this.

PR #416, opened 2016-02-06
Fix build of devel/readline on DragonFly.

ddbilik:
After privatization of ncurses, the system has no libtermcap to provide,
so remove TERMCAP from options.


jrmarino:
remove values for DEFINE_OPTIONS is often problematic.  If the only point of this is to remove a non-default option I think we should just do nothing.  There are probably hundreds of ports with bad non-default options (yes it's misleading, but we only guarantee that default options will build)

ddbilik:
OK, let's do nothing then. :) Thank you.

PR #434, opened 2016-04-05
Patch mgdiff

mneumann:


jrmarino:
we don't apply patches to patch files.

Normally we apply a second patch to the same file.
In extreme cases, we __REMOVE__ the freebsd patch file and replace it.

I can't approve the PR as it is now.

mneumann:
@jrmarino: thanks. this should work now.

jrmarino:
not quite.
1) It's in the wrong place.  You are aiming to make a patch for the dragonfly/ directory.  The patches in files/ get applied first, then dragonfly/.  You put it in diffs/ which only apply to the port itself, not the vendor software

2) I encourage you to use the ports-mgmt/genpatch tool to create patches like this.  You copied the original patch which is in a terrible format (starts with "./", wrong timestamp etc).  This is optional but that's what I do and obviously I prefer it.

jrmarino:
unrelated, but I want to catch you on on rust status.

1) lang/rust builds locally and will be in ports on next merge.  It is at version 1.7 and is required for firefox building (at least with default options).
2) lang/rust-nightly: The DF support is wrong.  I have added many patches to DeltaPorts and it builds for a while but coredumps with the stage2 rustc (version 1.9)
3) lang/rust-dragonfly: this is at version 1.3.  I will mark it "IGNORE" for now since lang/rust is a lot newer with the idea of removing the port soon.

ISSUE FOR 1) and 2)
Dragonfly is the only BSD that requires profcs to get the executable name.  I really want to change this.  For some reason, when I patch the file to use sysctl KERN_PROC_PATHNAME like freebsd, it just doesn't work.  I'd really, really like to fix it (if not I have to patch Synth and A) I don't want to and B) there's only one port that requires procfs and I'd rather it stay at 1 rather than move to 3) 

So I put a ton of work into Rust (something I don't use) so I'd like some help and ideally these patches need to get pushed upstream too.

mneumann:
@jrmarino thanks. 

Rust: I have patches to build lang/rust-nightly, but not lang/rust. If you made lang/rust work, that's perfect, because the Makefile in lang/rust-nightly depend on the other port. I have upstream patches, which I hope will be in 1.8, so rust builds out-of-the-box without any changes on Dragonfly. It coredumps, because a patch is missing which is related to jemalloc. Actually to build Rust you'd need https://github.com/rust-lang/rust/pull/32687 and https://github.com/rust-lang/rust/pull/32686.

Please remove rust-dragonfly.

procfs: I actually changed it from KERN_PROC_PATHNAME to use the procfs, because I couldn't get it to work otherwise on DragonFly. I think KERN_PROC_PATHNAME returns the resolved path, while procfs returns the virtual path, used when starting the executable. HAMMER had "@@" in the resolved path, I don't know if this caused problems or not. 

I can create a pull-request to get rust-nightly running. You might have to inspect it and do some minor changes.

mneumann:
@jrmarino : another try :)

jrmarino:
with the jemalloc fix, I assume I can get the current nightly to build.  For the errno, I just returned 0 because it seems only the sanitizers used errno_location.  I think the change was ok, but your errno TLS fix is better.

There were other ramifications to fixing the clock arguments though.  Maybe that's been fixed in head already.

We need to make another attempt at removing procfs.  It's deprecated on FreeBSD and should be in DF too.  There's only one port (before rust) that would not build without /proc mounted, rust will increase the number (when we want that number to be zero).

So in your opinion, the sysctl KERN_PROC_PATHNAME is returning a different format on DragonFly than FreeBSD?

mneumann:
@jrmarino: this should work for rust-nightly https://github.com/mneumann/DeltaPorts/commit/ba57680c7bd9b19cc8948876cf97f33d42f1f510

mneumann:
@jrmarino rust HEAD as of today contains all fixes. so the next version of rust-nightly will not need any fixes. that's why I kept all changes in a single file patch-dragonfly.

jrmarino:
if you create a separate pull request for rust-nightly I can merge it with one button and rebuild, thanks.

mneumann:
@jrmarino : pull request issued. I try figure out the problem with KERN_PROC_PATHNAME.

PR #435, opened 2016-04-09
Patch rust-nightly.

mneumann:
Probably needs testing. Worked within the freebsd-ports tree.


mneumann:
@jrmarino The difference between KERN_PROC_PATHNAME and procfs is as follows:
```
procfs: /home/mneumann/current_exe
sysctl: /data2/pfs/@@-1:00001/mneumann/current_exe
```

I think that the two "@" make trouble in path parsing or some other places. I am now building rust with KERN_PROC_PATHNAME and try to track it down.

mneumann:
@jrmarino  ok, figured out (and remembered) what the problem is. it's the colon in "@@-1:00001", and join_paths, which expects the string not to have a colon within it. I think this is a general problem with the way HAMMER encodes PFS in a path name. These pathnames can't be used in a PATH env variable, splitted by ":". Dunno what to do. Seems like escaping the ":" does not work in ```sh```.

jrmarino:
i'm wondering if that's the behavior we want for KERN_PROC_PATHNAME ...

mneumann:
yes. anyway, if we change the behavior of that ```sysctl```, we'd need to force anyone to upgrade or introduce version specific patches into dports.

jrmarino:
hmm, -nightly doesn't build:

```
configure: --enable-targets=x86,x86_64,arm,aarch64,mips,powerpc --enable-optimized --enable-assertions --disable-docs --enable-bindings=none --disable-terminfo --disable-zlib --disable-libffi --build=x86_64-unknown-dragonfly                         --host=x86_64-unknown-dragonfly --target=x86_64-unknown-dragonfly --with-python=/usr/local/bin/python2.7
./configure: /wrkdirs/lang/rust-nightly/rust-235d77457d/src/llvm/configure: not found
configure: error: LLVM configure failed
===>  Script "configure" failed unexpectedly.
```

jrmarino:
should 

```
.include "${MASTERDIR}/Makefile"
```

be the last line?  I don't know if defining stuff after that include is functional.

mneumann:
hm, I actually had those lines below the ```.include``` as a patch within the lang/rust/Makefile.

jrmarino:
right, that's what I'm saying.

Having them below means they don't work.  Usually the ".include" is the last line.

jrmarino:
it builds without those lines:
https://github.com/DragonFlyBSD/DPorts/commit/6475202cf04667590745e237bcbae38ef9e3155d

but devel/cargo no longer builds:
http://muscles.dragonflybsd.org/latest-builder-report/logs/errors/cargo-0.4.0.20150720.log

mneumann:
thanks for taking care of it!!!

cargo: Hm, ok, I think this is trying to build a very old version of cargo (0.5.0), while the one in freebsd-ports is a recent one (0.9.0).

The cargo port in freebsd-ports should work quite well. there is only a minor issue, when building with rust-nightly:

```
/usr/obj/dports/devel/cargo/.cargo/registry/src/github.com-0a35038f75765ae4/libgit2-sys-0.4.0/lib.rs:539:9: 539:30 error: unresolved import `git_status_show_t::*`. Not a module `git_status_show_t` [E0432]
/usr/obj/dports/devel/cargo/.cargo/registry/src/github.com-0a35038f75765ae4/libgit2-sys-0.4.0/lib.rs:539 pub use git_status_show_t::*;
```

A recent version of this crate (libgit2-sys) has a patch, that just removes the line that is responsible for this error. I don't know how to produce a patch that removes this line, but probably, this problem does only occur when compiling with rust-nightly (not with rust-1.7), which I do.

I am now trying to build rust-1.7 and see if I can build cargo with it.

jrmarino:
i had forgotten that cargo stopped building in January and it's been locked to the last version that built since then.  You should provide a pull request that completely updates deltaports devel/cargo including to remove the "LOCK" (change to "PORT"). thanks.

mneumann:
just build cargo successfully without any patches from freebsd-ports. https://github.com/DragonFlyBSD/DeltaPorts/pull/436

jrmarino:
Hi Michael,
lang/rust just got updated to v1.8.
I used the rust-nightly patches ( see https://github.com/DragonFlyBSD/DeltaPorts/tree/master/ports/lang/rust )
but I'm still getting a segfault.  Can you provide the missing info?  (lang/rust is critical for firefox)

mneumann:
seems like 1.8 was branched before my patches came in. The segfault is coming from the patch for jemalloc. I am building atm.

jrmarino:
add to the existing patches in deltaports lang/rust if you can.  I assume what's already in place is correct.

PR #437, opened 2016-04-18
Fix rust 1.8.0

mneumann:
liblibc was not up-to-date.


jrmarino:
the changes to Makefile.DragonFly don't look correct to me.

jrmarino:
also, why is src/libstd/rtdeps.rs patch removed?  we should have the same functionality as freebsd wrt execinfo.

mneumann:
hm, I tried directly from the freebsd port. All I had to do to make it work was to copy the ```patch-dfly-libc?``` patches and the ```patch-dfly_support```. Hm, it seems to work without ```rtdeps.rs``` patch. But I think it's better to keep this patch.

mneumann:
feel free to add back the options from Makefile.DragonFly. I haven't tested with those options.

jrmarino:
I was intending that the DeltaPorts version was "almost there" and only needed tweaks.  I didn't mean for you to start from scratch.
I know it builds without rtsdeps, but we want it to work right, not just build. :)

The Makefile.DragonFly stuff was critical, so unless you can defend why they were removed, we should leave it.

Finally, it sounds like you are building this stuff in a live environment (e.g. using make on modified /usr/dports)?  You really should be using Synth (or poudriere) to properly test these ports.  Are you familiar with ports-mgmt/synth?

mneumann:
yes, live environmet, no synth. I will use synth next time.

Hm, I am using rust since a long time and I always build it directly from it's repository. I **never** needed a patch like rtsdeps! And I run a lot of applications built with rust! So I strongly assume it works without this patch! But it does no harm either.

Yeah, the Makefile.DragonFly stuff was critical, that's why I was wondering why it built without it! IIRC, it did not build before without these modifications. I am not sure if building rust 1.8 with ```llvm``` from ports is the right thing, unless FreeBSD is doing the same. There are bug fixes in recent versions of llvm that Rust makes use of, which might not be in the llvm version in dports. The ```CFG_DISABLE_LIBCPP``` is only neccessary when building with ```clang```. But we build with ```gcc```. It does not harm either to keep it.

jrmarino:
"I never needed a patch like rtsdeps! "
That just means you didn't require backtraces I think.  If you had, it would have worked on FreeBSD but not DragonFly

"Yeah, the Makefile.DragonFly stuff was critical, that's why I was wondering why it built without it! " 
Probably you have llvm/clang installed and it's getting picked up, but they would not be there on a clean environment.

" I am not sure if building rust 1.8 with llvm from ports is the right thing, unless FreeBSD is doing the same. " 
We've got no choice, it doesn't build with base gcc

" The CFG_DISABLE_LIBCPP is only neccessary when building with clang"
right

"But we build with gcc."
incorrect, at least in my experience.
A clang-only flag is passed if CFG_DISABLE_LIBCPP isn't set.
IIRC the build fails later with GCC even if it is set.

AFAIK, ports rust has always been built with some kind of clang.

jrmarino:
ah I just remembered, I had to modify synth to build rust because of the procfs thing.  It will not work for you unless you have the same modification (or rust is fixed to not use procfs)

mneumann:
hm, backtraces work for me. maybe execinfo is linked in at some other place. I know that FreeBSD always builds rust with clang, but when I compiled rust from repo, I always used gcc.

if you figure out which flags and patches we need to compile rust 1.8, feel free to commit. 

"procfs": I think we need dillon for some suggestions on this topic.

jrmarino:
I'll probably have to go ahead and push the synth modification because it sounds like the solution to procfs will be a while.  I should write a little program to see the output of KERN_PROC_PATHNAME (or whatever it is) sysctl for both HAMMER and UFS file systems and what it looks like on FreeBSD's ZFS and UFS.  I would not be surprised if our sysctl's output isn't what we intended.

jrmarino:
oh geez, at least part of my problem is that I wasn't using the modified synth to build rust.  Now I'm getting an old problem back (using your 4 new liblibc patchs + the dfly_support patch + the execinfo patch):

```
src/libstd/sys/unix/net.rs:24:1: 24:31 warning: `pub extern crate` does not work as expected and should not be used. Likely to
src/libstd/sys/unix/net.rs:24 pub extern crate libc as netc;
                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/libstd/sys/unix/time.rs:245:40: 245:61 error: mismatched types:
 expected `i32`,
    found `u64` [E0308]
src/libstd/sys/unix/time.rs:245             Instant { t: Timespec::now(libc::CLOCK_MONOTONIC) }
                                                                       ^~~~~~~~~~~~~~~~~~~~~
src/libstd/sys/unix/time.rs:245:40: 245:61 help: run `rustc --explain E0308` to see a detailed explanation
s
```

I'm wondering how I'm getting this and you aren't?
Are you building on DF master or like release 4.4?

jrmarino:
I pushed your patches (as one big one) and removed the existing liblibc patch here: https://github.com/DragonFlyBSD/DeltaPorts/commit/ecdd0db7b9ac66125a67f460da145a9cf9a3a43c

jrmarino:
Also, I updated synth to version 1.34 and that version will mount procfs so that it can build rust and rust-nightly.  Hopefully that's a temporary thing. :)

mneumann:
this should fix the compile problems: https://github.com/DragonFlyBSD/DeltaPorts/pull/439. thanks for all the work you are pushing into this!

jrmarino:
This version built for me locally, successfully.  thanks!

jrmarino:
FYI, this procfs thing is getting serious.
We can't build firefox with rust options on in Synth anymore

```
error: internal compiler error: unexpected panic
note: the compiler unexpectedly panicked. this is a bug.
note: we would appreciate a bug report: https://github.com/rust-lang/rust/blob/master/CONTRIBUTING.md#bug-reports
note: run with `RUST_BACKTRACE=1` for a backtrace
thread 'rustc' panicked at 'can't determine value for sysroot', src/librustc/session/filesearch.rs:165
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

can we use progname or something in the short term?

jrmarino:
There's another trick we can try:
I can add a modification to the port via Makefile.DragonFly to write the absolution location of the rustc executable in a string in the sourcefile, and the function just returns the string.

e.g., we know the localbase (defaults to /usr/local) so the location would be "/usr/local/bin/rustc" so we can just return a static value.

right?

mneumann:
hm, this only works when called from ```rustc```. There might be other binaries (rustdoc, cargo) calling this function.

```__progname``` would probably work *iff* we can access it easily from Rust. I don't have DragonFly at hand for the next days, so I can't try it atm.

jrmarino:
"There might be other binaries (rustdoc, cargo) calling this function."
is it hard to check? I agree if in doubt, we can't use that technique.

The solaris handling works off progname or something similar IIRC.  I remember even thinking we could do the similar implementation.  It's worth looking at.

I was just thinking of a better provisional way that procfs, with the idea of using sysctl being the ultimate goal (although experience says if the provisional way works well enough, it will never be changed)

zrj-rimwis:
@mneumann is https://leaf.dragonflybsd.org/~zrj/patch-src_libstd_sys_unix_os.rs correct?
Allows to build www/firefox in synth.
My quick tests on hammer show that both sysctl and /proc (C program) return identical paths with "@@". So I'm not sure why /proc would be more preferable than sysctl.
Unless I'm missing something else,  like it using that string without escaping special symbols in environment variables (like PATH)

mneumann:
we found a solution :)

jrmarino:
Hi mneumann,
I think I mentioned it before, but just in case: The dports generation is blocked by your upcoming rust update.  It's a hard block.  That's the reason dports hasn't been generating in the last few weeks.

Do you have an ETA for your rust work so we can start the builds again?

PR #447, opened 2016-05-06
x11-drivers/xf86-video-ati: Newport it to update.

zrj-rimwis:
Current oficial 7.7 version performs better on xorg v1.17.4.
Also, contains both new supported hardware (like R7 360 bonaire)
and driver fixes that are hard to backport back to 7.5 version.

Tested-on: master w/ JUNIPER, CAICOS, OLAND and BONAIRE cards.


jrmarino:
fyi, you should have set the PORTREVISION to 0 (remove it) since the version number increased.

zrj-rimwis:
bah, that one slipped through..
Thanks for noticing and fixing it

PR #449, opened 2016-05-19
Fix opensmtpd-devel

mneumann:


jrmarino:
i think you should change "pre-configure:" target to "dfly-patch:".
We have our own target now that applies after "post-patch:" target.
(We used to use pre-configure like you just did before we had it).  What do you think?

jrmarino:
I'll fix this locally.

jrmarino:
ugh, diff failed.  it's a bad patch

mneumann:
I've taken the patch from ```opensmtpd``` port. @jrmarino : I know how to run ```synth``` to test a port from ```dports```. But what's the procedure for DeltaPorts? Is there an easy way to produce a port from a DeltaPorts description? Or do I have to fully install the DeltaPorts system (in the docs it's still using poudriere)?

jrmarino:
the patches have to be generated against the same port, e.g. mail/opensmtpd-devel, not mail/opensmtpd.

You can use https://github.com/DragonFlyBSD/DeltaPorts/blob/master/scripts/generator/sync1.sh to generate a "potential" port, you need to define the locations of freebsd ports and deltaports and the target directory.  Just review the generator directory and I'm sure you can figure out how to define dports.conf.

jrmarino:
You could make the target directory /usr/dports I suppose and it would dump it there, which may be the best for you.

mneumann:
thanks!

PR #450, opened 2016-05-20
Fix pty-related code of asciinema for DragonFly

mneumann:
patches also sent upstream.


jrmarino:
is, for example, patch-util_select.go functional?  those look like comments, but do those comments do anything?  If not, I can see fixing them upstream but not in the port.

mneumann:
yes, it's functional!!! ```+build ...``` are build instructions, and tell the go compiler to compile this file only on those platforms. They are absolutely mandantory!!! Go has many ways to compile platform-specific code. ```+build``` is one way. You also see files ending in ```_freebsd``` or ```_dragonfly```. This is similar to a ```// +build freebsd``` instruction.

PR #451, opened 2016-05-22
Fix for DragonFly

mneumann:


jrmarino:
technically /usr/local shouldn't be hardcoded, but it's okay for 99% of the use cases

mneumann:
yes, I know. just copied what was in there for FreeBSD

jrmarino:
right, I mean the port was already wrong before you got to it (and the maintainer if there is one needs to fix it) but that issue is out of scope of this one.

mneumann:
@jrmarino FYI, this fix is to use [GMIC](http://gmic.eu/) video effects for [FlowBlade](http://jliljebl.github.io/flowblade/). Flowblade is a nice video cut program. There is no port for it, but it runs from the repo once you install ```pkg ins py-mlt py-pillow py-dbus gimp-gmic-plugin```.

PR #452, opened 2016-05-22
Fix a longstanding issue with named pipes in libuv

mneumann:
This is a [bug][1] in libuv. The type of a named pipe was not correctly
identified on DragonFly, which led to the following problems in node:

    internal/process/stdio.js:152
          throw new Error('Implement me. Unknown stream file type!');
          ^

    Error: Implement me. Unknown stream file type!
        at createWritableStdioStream (internal/process/stdio.js:152:13)
        at process.stderr (internal/process/stdio.js:25:14)
        at Socket._destroy (net.js:476:25)
        at Socket.destroy (net.js:510:8)
        at maybeDestroy (net.js:434:12)
        at Pipe.onread (net.js:562:5)

[1]: https://github.com/libuv/libuv/pull/884

### DIFF
--- a/ports/www/node/dragonfly/patch-deps_uv_src_unix_stream.c
+++ b/ports/www/node/dragonfly/patch-deps_uv_src_unix_stream.c
@@ -1,0 +1,17 @@
+--- deps/uv/src/unix/stream.c.orig	2016-05-22 22:43:49 UTC
++++ deps/uv/src/unix/stream.c
+@@ -961,6 +961,14 @@ uv_handle_type uv__handle_type(int fd) {
+     return UV_UNKNOWN_HANDLE;
+
+   if (type == SOCK_STREAM) {
++#if defined(_AIX) || defined(__DragonFly__)
++    /* on AIX/DragonFly the getsockname call returns an empty sa structure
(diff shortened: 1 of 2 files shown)


jrmarino:
bad patch:

1 out of 1 hunks failed--saving rejects to deps/uv/src/unix/stream.c.rej
=> Patch patch-deps_uv_src_unix_stream.c failed to apply cleanly.
=> Patch(es) patch-deps_openssl_config_opensslconf.h applied cleanly.
*** Error code 1

mneumann:
ah ok, DeltaPorts has node 6.2.0, while I tested the patches with the version 5.11.0 in dports. will come up tomorrow with an updated patch.

jrmarino:
you *have* to generate patches against FreeBSD ports, not DPorts.   Doing the latter is an error.  I think I even recall this happening before.

mneumann:
Doh! It probably happend before :(. Now this should correct it: https://github.com/DragonFlyBSD/DeltaPorts/pull/454. It is based on freebsd ports, with the ```dragonfly/``` directory copied into it. I was able to successfull install node 6.2.0 with it.

PR #460, opened 2016-06-09
graphics/weston: Add --with-xserver-path configure option for xwayland.

ivadasz:


jrmarino:
will anything break if "${PREFIX}/bin/Xwayland" does not exist?

ivadasz:
no, nothing will break. weston will just fail to launch Xwayland (which it tries to do the first time the user tries to run an X application from the weston session). But weston will just continue running despite failing to launch Xwayland.

jrmarino:
cool.  I didn't realize this was a newport until now anyway.  approved.

PR #462, opened 2016-06-11
www/npm: Remove obsolete patch

mneumann:
Since the acceptance of the patch to libuv introduced in https://github.com/DragonFlyBSD/DeltaPorts/pull/452 the module uidNumber works with nodejs. As a result, the patch introduced in https://github.com/DragonFlyBSD/DeltaPorts/pull/1 is no longer required.

### DIFF
--- a/ports/www/npm/Makefile.DragonFly
+++ b/ports/www/npm/Makefile.DragonFly
@@ -1,3 +1,0 @@
-dfly-patch:
-	${REINPLACE_CMD} -e "s|'nobody'|'${:!/usr/bin/id -u nobody!}'|" \
-		${WRKSRC}/lib/config/defaults.js


jrmarino:
I don't know what's going on here.
At one time, this change to www/npm was working, but now it's not.  Returning the old MD allows it to build again.  It's like www/node doesn't have the fix anymore.  Please investigate.

mneumann:
@jrmarino I think the reason why it fails is because of this change https://github.com/DragonFlyBSD/DPorts/commit/43bf3899f992818bbbf182ca426bfe8d34172eb2, which builds nodejs with a shared libuv, i.e. it depends on devel/libuv now. Not sure if we have the same patches applied there yet. I will take a look. Btw, this means, we should be able to remove the libuv patches from www/node.

PR #468, opened 2016-06-15
www/firefox: Add slave port for lite package

zrj-rimwis:
Intended to avoid dep on gtk3 and ensure gtk2 one still functional.

Build order:
  first rebuild www/firefox to include updated MD
  then in separate build, attempt the www/firefox-lite slave


zrj-rimwis:
LITE handling got into delta but not in dports, so www/firefox-lite just rebuilt www/firefox with same txz name.
Could repeat www/firefox to update MD in dports too

jrmarino:
hmm?  I did rebuild www/firefox.  It took 30 minutes.

PR #515, opened 2016-06-16
devel/acsccid: Fix MF for missing OPSYS

zrj-rimwis:


jrmarino:
this one I should probably push upstream
(as well as any USES+=ncurses)

PR #558, opened 2016-06-25
www/chromium: Attempt to allow build new version

zrj-rimwis:
Lots of shifting around, but it builds for synth just-build
synth test fails because activates TEST option and build fails on linux testing sources
dunno what is missing for those

Just build fix, untested runtime! archived version at ~zrj/dports/chromium_try.tar.bz2

btw i just pushed some stuff directly, and it didn't got into last run (rsync?)

### DIFF
--- a/ports/www/chromium/Makefile.DragonFly
+++ b/ports/www/chromium/Makefile.DragonFly
@@ -2,3 +2,37 @@
 .if exists(/usr/include/bus/u4b/usb_freebsd.h)
 CFLAGS+=	-D__STDC_LIMIT_MACROS
 .endif
+
+KERBEROS_LIB_DEPENDS+=	libgssapi_krb5.so:security/krb5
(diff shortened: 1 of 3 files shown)


jrmarino:
1) excellent timing, I was about to look at this
2) I don't see a push to www/chromium directory
3) It would have still failed because it would have to be synced manually and I wasn't aware that was needed
4) as of right now, I suspect the TEST issue is resolved.  I have been trying to get FreeBSD to revert that shit but it's not moving, so I removed it manually from bsd.port.mk on this cycle.  It's in dports for the last few seconds.

zrj-rimwis:
i pushed only fixes mainly to sysutils/ and audio/
chromium is a bit more complicated, builds > 3h+..

jrmarino:
yeah, ccache is a must for chromium

PR #559, opened 2016-06-25
lang/pcc: Hook pcc-libs

zrj-rimwis:
Since they finaly were added to freebsd-ports, now include dragonfly too.
Simple hello world .c works fine.


zrj-rimwis:
btw i have touched bsd.df.gcc.mk.diff (removed ${PORTSDIR} from origins)

PR #560, opened 2016-06-27
multimedia/libmediaart: Fix plist

zrj-rimwis:
Noticed that fails in synth test


jrmarino:
what about this?
https://svnweb.freebsd.org/ports/head/multimedia/libmediaart/pkg-plist?r1=393607&r2=416822

jrmarino:
see: https://lists.freebsd.org/pipermail/svn-ports-all/2016-June/127045.html

zrj-rimwis:
yeah, not sure about this one, almost was tempted to do post-install: rm *.sgml
but it looks like it should be there based on index.html etc
p.s. was trying to fix sysutils/catfish that takes it as a dep

jrmarino:
accepted provisionally.  It will surely be reverted soon since it's a mistake for FreeBSD.

PR #564, opened 2016-06-29
textproc/docbook-xsl-ns: Don't package .tmp files

zrj-rimwis:
silly one


jrmarino:
I pushed an equivalent fix upstream

PR #565, opened 2016-06-29
multimedia/libva: Fix build on new version

zrj-rimwis:
Drop alias and patch accordingly (port now contains sys dep stuff)
there is an issue with libEGL (when it is present, extra backend gets enabled, plist stuff)

Only merge just before next bulk (patch is just to reduce fail counts on next)


jrmarino:
I pushed equivalent fix upstream

PR #575, opened 2016-07-12
devel/libqb: fix compilation

dkgroot:


jrmarino:
one doesn't redefine USES, they use USES+= 

e.g. USES+= autoreconf

zrj-rimwis:
Thanks for the patch, pushed slightly reduced version. We prefer to avoid autoreconf when possible.

dkgroot:
Sorry about that, didn't know. Thanks for fixing and adding it !

PR #576, opened 2016-07-12
net/corosync: ported from FreeBSD

dkgroot:
Depends on previous PR for devel/libqb


zrj-rimwis:
closing this one as net/corosync should automatically be added in the next bulk build.
Note however, quick test by running /usr/local/sbin/corosync causes segfault.

dkgroot:
I didn't notice the segfault during my tests. Will recheck and try to find the cause once it is merged.

PR #577, opened 2016-07-14
fix recent broken patches for CRLF

zrj-rimwis:
Use DOS2UNIX


zrj-rimwis:
correct? should i revert previous repush of LF->CRLF in patches and change to use DOS2UNIX too?

also try building misc/seabios (gcc4.8 and forced lf.bfd otherwise doesn't build)

zrj-rimwis:
there are more dragonfly patches that has CRLF: http://sprunge.us/LciK
so leave them for now, or convert to DOS2UNIX_FILES ?

jrmarino:
i prefer DOS2UNIX always, so it's a good change as far as I am concerned

jrmarino:
i would leave anything that works and change it only when the patches need to be regenerated.  no need to do extra work.

PR #579, opened 2016-07-16
Fix a longstanding issue with named pipes in libuv

mneumann:
These are the same patches as those found in [1] for the shipped version
of libuv of nodejs. Recent port revisions of nodejs (www/node) start to
build with the libuv version from dports. This led to a regression of
this bug in nodejs (and npm), as we haven't fixed this in devel/libuv
(yet). This commit fixes this issue.

[1]: https://github.com/DragonFlyBSD/DeltaPorts/pull/452


mneumann:
@jrmarino Once this lands, you should be able to remove the patches to the shipped libuv in www/node as introduced by https://github.com/DragonFlyBSD/DeltaPorts/pull/452. And you can also reapply the patch from https://github.com/DragonFlyBSD/DeltaPorts/pull/462.

mneumann:
@jrmarino: Just tested what I described above, and it works here.

PR #581, opened 2016-07-19
devel/gdb: Fix build after new patches

zrj-rimwis:
It builds in synth and runs on simple quick small programs.
Just one thing: on "layout split && break main && run" sometimes gdb segfaults.
Seems like ncurses issue, but mostly it works.

DragonFly support in gdb should be really be fixed and upstreamed...


jrmarino:
agreed but nobody has submitted the paperwork. I'm probably the most close to be able to do so though.

PR #582, opened 2016-07-19
sysutils/e2fsprogs: Fix build after enabled tests

zrj-rimwis:
For now just disable m_hugefile test, no idea why it would fail
also expand patching

Update: disable d_dumpe2fs_group_only, fails on synth test...
also leaves /etc/blkid.tab behind in base..


zrj-rimwis:
it is likely to fail on muscles too, so we might need to disable running tests all together

jrmarino:
there was already a change upstream to a couple of these ports, specifically to fix the -luuid error.  I suspect some of these changes are redundant now.

jrmarino:
okay, they are probably fine.

PR #589, opened 2016-07-21
devel/commoncpp: Convert alias to patches

zrj-rimwis:
commoncpp installs common headers that are used by ports depending on it.
Since it building CFLAGS aren't fully considered everywhere it is better
to patch such ports directly to provide fixed headers than depend on
implicit alias + strange patches in other ports.

This fixes another case of upcoming POSIX ns cleanup in base.
Specifically net/panoptis that for now passes w/o any patching.

To rebuild: audio/ccaudio2 devel/commoncpp net/panoptis

Disected-by: swildner

### DIFF
--- a/ports/audio/ccaudio2/Makefile.DragonFly
+++ b/ports/audio/ccaudio2/Makefile.DragonFly
@@ -1,1 +1,0 @@
-USES+= alias:11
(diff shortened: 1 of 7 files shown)


zrj-rimwis:
Maybe we need to add special target in pre-package that checks for \__FreeBSD__ in staged/*.h? Not that it should abort, just to write in build log smth like "DFLY_WARNING: installed header with bare \__FreeBSD__" and maybe grep -n3 on that file for later analysis
would allow to reduce patched port counts

PR #598, opened 2016-08-03
sysutils/flock-dfly: Bandage

zrj-rimwis:
We are missing POSIX timer_* functions and new sysutils/flock needs them.
For now only chromium and google-startup-scripts have a dep on flock
(except for www/mod_spdy that is still in failing heap).

Keep this port until we implement the missing POSIX functions.

### DIFF
--- a/ports/net/google-startup-scripts/Makefile.DragonFly
+++ b/ports/net/google-startup-scripts/Makefile.DragonFly
@@ -1,0 +1,3 @@
+
+# zrj: sigh.. until we grow timer_create und friends..
+RUN_DEPENDS:= ${RUN_DEPENDS:Nflock*} flock:sysutils/flock-dfly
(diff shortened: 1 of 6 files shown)


jrmarino:
I don't think I agree with this.
why not just set LOCK in STATUS to keep the existing 2.19 in place?

zrj-rimwis:
We need to implement posix timer_* functions in base or other ports will start to break too.
Also looks like previous maintainer zont owns the primary MASTER_SITES www.zonov.org
thus separate port with fallback to my leaf ~zrj distcache is safer until we can deal with timer_* stuff on master.

jrmarino:
i'm not arguing against adding timer_* functions, I'm just saying that in this particular case, it seems better to LOCK down the port (as long as chromium still builds with it).

Once we have it in base and the last release, we can remove the LOCK.

jrmarino:
i locked flock.  let's implement the timer function now and unlock flock at the next release.

PR #599, opened 2016-08-03
ftp/R-cran-curl: Fix build on new version.

zrj-rimwis:
In ports FreeBSD had a general patch, yet upstreamed just their case.
Patch accordingly for DragonFly too.

Btw, shouldn't main math/R port be somewhat more like: http://sprunge.us/XWOT


jrmarino:
blet ??

jrmarino:
https://svnweb.freebsd.org/ports/head/math/R/Makefile?r1=419530&r2=419529&pathrev=419530

zrj-rimwis:
oh that should "exec" shouldn't have entered into that sprunge..

PR #602, opened 2016-08-05
net/libpcap: Fix build w/o IPV6

zrj-rimwis:
How this could have passed through all quality checks?

Allows to build my package set w/o IPV6 option set.


jrmarino:
DF only?  does freebsd need this too?

zrj-rimwis:
It looks like they should need it too, strange that this is in an official libpcap release.

PR #613, opened 2016-08-17
games/iortcw: Add DragonFly

zrj-rimwis:
User request for irc.
Game requires pak0.pk3 to run from RTWC original CD that I don't own.


zrj-rimwis:
Just moments pushed to freebsd-ports, so should be ok for next incremental.
Game exe is functional just doesn't go into GL mode w/o .pk3 files.

PR #623, opened 2016-08-18
misc/qt5-doc: Wild guess.

zrj-rimwis:
dunno what is wrong in pou, swildner reported similar build failures,
yet on synth test no such problems so far.
Try using gmake, it gives less warngins on generated makefiles.
Also add make jobs unsage in MD too, maybe pou gets confused for it after
bsd.port.pre.mk


jrmarino:
i'd rather leave qt5-doc ignored.  I already removed it from qt5 meta port.  The only other dependency is qtcreator which never built for other reasons.

qt5-doc has always been flakey and fragile, so I'd rather not fool with it anymore.

PR #631, opened 2016-09-07
security/openssl-devel/: Fix build + is there a bug?

zrj-rimwis:
Looks like there is a bug in crypto/engine/eng_cryptodev.c from 1.0.2h -> 1.1.0...

synth test PASS
gmake test gives
All tests successful.
Files=85, Tests=453, 51 wallclock secs ( 0.36 usr  0.20 sys + 29.55 cusr 21.40 csys = 51.52 CPU)
Result: PASS


zrj-rimwis:
seems to be this upstream commit https://github.com/openssl/openssl/commit/b61d2da71b246fbc24af0dec2b15dc49141d5990

PR #633, opened 2016-09-16
sysutils/fusefs-libs: Initial compat fixes.

zrj-rimwis:
Will depend heavily on kernel side implementation.

Requested-by: tkusumi@


jrmarino:
i don't understand why this is here now.
Until there's actually a libfuse implementation, I can't merge this.  (Unless I'm missing something).

PR #636, opened 2016-10-03
editors/emacs: Fix build w/ gcc50 and C-x C-f browser.

zrj-rimwis:
Prevent calloc() optimization (just as for csh in base).

Also fix reports by emacs users that file viewer fails on special filenames.
Turns out emacs is very relying on ls having --dired flag support (gnu ls).
So if it fails it fallbacks to regexes on ls -l output that results in some
unselectable file entries.
Fix it by adding new option that uses gls from coreutils cause even swildner
doesn't like idea of adding --dired support in base ls. From gls(1):
       -D, --dired
              generate output designed for Emacs' dired mode
In case of GNULS=off, base ls is used, it works for standard filenames
(no whitespaces, special symbols or complicated symblinks --> ../../bin/../file)
so just enforce LC_TIME=C for ls -l invoke to ease up output for dodgy regexes.

### DIFF
--- a/ports/editors/emacs/Makefile.DragonFly
+++ b/ports/editors/emacs/Makefile.DragonFly
@@ -1,1 +1,10 @@
-USE_GCC=	NOT5
+#USE_GCC=	NOT5
+CFLAGS+= -fno-builtin-malloc # yay it is still needed :D (calloc opt)
+
+OPTIONS_DEFINE+=	GNULS
+OPTIONS_DEFAULT+=	GNULS
+
+GNULS_DESC=		Use gnu ls 'gls' for --dired flag support
+GNULS_EXTRA_PATCHES_OFF=	${FILESDIR}/../dragonfly/extra-ls_lisp_dired.el
+GNULS_EXTRA_PATCHES=		${FILESDIR}/../dragonfly/extra-gls_lisp_dired.el
+GNULS_RUN_DEPENDS=		gls:sysutils/coreutils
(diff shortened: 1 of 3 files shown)


jrmarino:
The patch itself looks fine but is not FreeBSD affected by this "dired" issue?

jrmarino:
also, what about emacs-devel?

zrj-rimwis:
Hard to say, for now lets play it safe and give users time to test it, likely not so many uses in-built file navigator so they don't see the issue.
This patch was tested by tuxilo.
As for emacs-devel, it looks like it properly builds with gcc50

zrj-rimwis:
We don't see the use of what --dired does, annotates where filename starts on each line and thus emacs can avoid to use regexes on output when gnu ls is used.
ls -la --dired /
http://sprunge.us/DhXa

jrmarino:
I mean, doesn't emacs-devel also require dired handling?

zrj-rimwis:
if it has same logic in lisp/dired.el then yes, but i doubt we should patch it too since users are not likely using devel package since it contains Xorg support, so I *guess* most have emacs-nox11 lite package.

jrmarino:
The way I see it, if -devel has the lisp/dired.el, it needs the same treatment.  We can't pardon a known problem by saying we don't think people are using the editor.  IMO

zrj-rimwis:
devel version is still at rc2 (v25.1) maybe first lets wait for reports if using gls doesn't cause other issues and once v25 (that doesn't need -fno-builtin-malloc) replaces current v24 we could add GNULS option to both.

jrmarino:
well, I'd like to be more proactive with emacs-devel.  It could replace editors/emacs at any time (although as long as it's implemented in Makefile.DragonFly we'd be okay).

If we confirm this is the right fix, we need to pass it to FreeBSD and it needs to be a mandatory dependency (who would choose to have a broken editor?)

PR #650, opened 2016-11-11
mail/spamd: Attempt to unbreak build.

zrj-rimwis:
Untested, other than synth test.

Requested-by: Studbolt


jrmarino:
honestly, the maintainers of spamd should update it to use a newer hash.  This level of "fixing" pushes the limits of what dports maintainers should be doing, IMO.

PR #651, opened 2016-11-13
Fix cargo 0.13

mneumann:
This also uses my statically linked cargo binary.


jrmarino:
hmmm, interesting that the name itself is significant.
I had already done most of this work, but I took the missing data from your PR.
see c5eba56cf073f424599e147efdca9cfaf4fd68fe

thanks!

PR #669, opened 2016-12-13
devel/kore: Remove clangism

zrj-rimwis:


jrmarino:
this one should be fixed soon anyway

PR #671, opened 2016-12-13
sysutils/pick: Fix ncurses dectection.

zrj-rimwis:


jrmarino:
i should probably push this one (or equivalent) to ports ...

PR #679, opened 2016-12-13
games/openbor: Fix build after ports update.

zrj-rimwis:
FreeBSD removed the patch for max_align_t, since we don't want to add
such non standard additions to our standard libs just patch it directly.
While there use correct order for linking libs.


jrmarino:
agreed, I already told Jan we weren't touching stddef.h, so we are in agreement.

PR #697, opened 2016-12-17
converters/rubygem-bsdconv: Unbreak package.

zrj-rimwis:
Is it broken on freebsd-ports too?


jrmarino:
removed from dports because last commit on ports broke it.

PR #712, opened 2016-12-17
audio/murmur: Unbreak build

zrj-rimwis:
c++: error: unrecognized command line option '-Wl', how that even works w/ clang?
Synth test only (dbus/qt)


jrmarino:
pushed upstream. 
The extra patches isn't worth fixing because A) it never applies and B) it will be removed entirely in 2 weeks

PR #715, opened 2016-12-17
binutils 2.27 ld.gold workarounds pt2

zrj-rimwis:
Explicitly link with libm to allow building rest of Haskell ports.

Before incremental remove these 4 binary txz:
hs-hslua-*.txz hs-JuicyPixels-*.txz hs-dia-functions-*.txz hs-mwc-random-*.txz


jrmarino:
We still have one failure:
devel/hs-git-annex

PR #716, opened 2016-12-18
devel/subversion: Do not provide freebsd template by default.

zrj-rimwis:
User request: the way template is contructed causes issues with python tools.

I didn't asked for more details, but given that files/extra-patch-fbsd-template
performs direct subversion code modification it is better to ship original.

### DIFF
--- a/ports/devel/subversion/diffs/Makefile.diff
+++ b/ports/devel/subversion/diffs/Makefile.diff
@@ -1,0 +1,11 @@
+--- /data/freebsd-ports/devel/subversion/Makefile	2016-11-17 15:56:03.000000000 +0200
++++ Makefile	2016-12-18 13:29:10.000000000 +0200
+@@ -26,7 +26,7 @@ OPTIONS_DEFINE=	\
+ 		TEST		\
+ 		TOOLS
+ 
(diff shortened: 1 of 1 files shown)


jrmarino:
why a Makefile.diff instead of Makefile.DragonFly ?

zrj-rimwis:
there are slave ports

zrj-rimwis:
or even better:
.if ${OPSYS} == FreeBSD
OPTIONS_DEFAULT+= FREEBSD_TEMPLATE
.endif

jrmarino:
OPSYS isn't available for options_default evaluation.
Even slave ports pull in the top local makefile I think.

zrj-rimwis:
wouldn't OPTIONS_DEFAULT_FreeBSD+= FREEBSD_TEMPLATE be enough
there are OPTIONS_EXCLUDE_FreeBSD_10+= DTRACE

jrmarino:
you mean to change upstream at the ports level?

zrj-rimwis:
yup

jrmarino:
there's no such construct as OPTIONS_DEFAULT_${OPSYS}

PR #735, opened 2017-01-09
security/cyrus-sasl2-gssapi: Work around messy heimdal.

zrj-rimwis:
I think it is just better to use GSSAPI_MIT here.
Nothing depends on this lib.


jrmarino:
".if 1" evaluates to zero?  
Isn't that kind of misleading?

zrj-rimwis:
i think it works as expecting .if 1 - > true, .if 0 -> false

jrmarino:
then the commit doesn't match the description.  You've kept heimdal, not switched to MIT

jrmarino:
i changed the option manually, PR is now obsolete.
Oh, I didn't realize I closed the PR already.  I wanted to close it now though.

PR #771, opened 2017-02-13
devel/linux-kernel: Prefer to build x86_64.

zrj-rimwis:
There is something strange, this fails to build with lang/gcc6-aux
over on video_mode.c -m16 with reg not being multiple of 8.
For now just use gcc 4.9.

### DIFF
--- a/ports/devel/linux-kernel/Makefile.DragonFly
+++ b/ports/devel/linux-kernel/Makefile.DragonFly
@@ -1,0 +1,6 @@
+
+# undo freebsd'isms, we want 64-bit version, not 32-bit one
+dfly-patch:
+	${REINPLACE_CMD} -e 's@amd64@x86_64@g'	\
+		${WRKSRC}/Makefile \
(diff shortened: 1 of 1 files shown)


jrmarino:
That sounds like the same problem gprolog had.
Can't we fix it the same way?

Do we actually need this port at all?

zrj-rimwis:
Hmm not sure, gprolog seems to be building with gcc50 based on logs.

jrmarino:
because I fixed gprolog so I could remove USES_GCC=NOT5

jrmarino:
ah, it wasn't gprolog, hold on.

jrmarino:
it was misc/seabios, sorry.

zrj-rimwis:
i thought something is off with gprolog, yeah seabios might be it.
So just -m16 --> --param large-stack-frame=4,0,-fno-inline ?

as for building with base gcc50 we could try this http://sprunge.us/Uace

PR #774, opened 2017-02-16
D1702 try mesa update

zrj-rimwis:
i915 seems to be functional radeonkms is not (fails to DRI2 authenticate)


zrj-rimwis:
too unstable

PR #778, opened 2017-02-23
ports-mgmt/poudriere-devel: Change default DPorts.git location.

zrj-rimwis:
swildner reported some strange differences between pou and direct from ports
compilation in logs like.
While comparing build logs i noticed that devel/m4 is compiled with
LIBSIGSEGV=on while dports not.
Turns out it is fetched by poudriere ports -c


jrmarino:
good catch.  my fork is all forked up (it has a special purpose).  it can't be used.

PR #780, opened 2017-04-24
htop: Added dragonflybsd implementation for htop

dkgroot:


jrmarino:
I'm not sure what I'm looking at.
It's neither an overlay to freebsd's sysutils/htop (preferred) nor a "newport" a newport.
If I merged this PR it wouldn't build (in fact it would break the tree).

dkgroot:
There was already an htop directory in Deltaports with a status file (failing). I guess it used to be based on a (much) older version of htop using (Linux) procfs. When dragonflybsd still had procfs support, htop would work if procfs was installed. Htop went from /proc (Linux) only to supporting multiple operatingsystems natively.

The files I committed are based on the current FreeBSD htop Ports files (so what you called an overlay).
What should I change (or have done differently) to make this work ?

dkgroot:
Note: I have also sent a pull request to the htop repository to include the dragonflybsd specific files [See](https://github.com/hishamhm/htop/pull/613). The maintainer of the htop repo has quite a bit of backlog though, so it will take a while for the patch to be accepted (I think). Once the pull request does get accepted and a new version is released I will update the Deltaports to match.

jrmarino:
deltaports is the difference between a modified port for dragonfly and the reference port in the freebsd ports collection.  Generally this difference comes in the form of additional patches or an additional makefile, and occasionally a modification to the port files themselves.

i think what you updated was a diff to the vendor sources.

I would suggest copying over freebsd port of htop to dports, and then git it to work there.  then show a list of differences.  Basically you're going to need to generate a lot of patches to go into the "dragonfly" patch directory.

dkgroot:
I copied the FreeBSD Ports Version to /use/dports and worked on it to get it to compile cleanly. I then copied the result to the Deltaports directory.  Not 100% where I went wrong to be honest.

jrmarino:
hmm, maybe I misread it the first time. it looks okay when I look at it again.
I'll merge it and see if it builds.

jrmarino:
it built fine and is now published in dports.  Thanks!

dkgroot:
Ooh ok nice. Thanks for taking to time to try again ☺

PR #785, opened 2017-07-03
Fix rust

mneumann:


jrmarino:
rust builds now.  The tweak to MD was added independently so this PR won't apply and isn't needed.

PR #786, opened 2017-07-04
Change PTHREAD_STACK_MIN

mneumann:
This should prevent cargo from failing when running with multiple
processes (not tested).

http://gitweb.dragonflybsd.org/dragonfly.git/commitdiff/592912cde1dc358bf080ae9aff3eca491688f47c

https://github.com/rust-lang/libc/pull/643


jrmarino:
actually I accepted this too fast.
This has to be a conditional patch. 
It can only be applied to head of current, right?  Changing this value for older platforms is wrong, isn't it?

mneumann:
I don't think this is wrong on older platforms. It is only the minimum stack size needed to run a thread, and the 1k was actually wrong, as the threads just imploded in rtld. This fix would fix nodejs on older platforms as well.

PR #788, opened 2017-07-06
New rust snapshots without pthread issues

mneumann:


mneumann:
argh!!! I am seeing hangs again on my box with these snapshots! Not 100% if the pthread issue did really fix it or if there are other issues! I am building again...

mneumann:
@jrmarino : Please revert this commit :(. It's hanging on my box.

jrmarino:
git is not co-operating with the revert command.
Plus I am a little concerned that the distfiles are not changing name, but hashes.
That means the original bootstrap files have already been overwritten.
In general, new distfiles should have unique names.

i might have to apply a reverse patch, but only after we've coordinated to make sure the bootstrap files are available.

jrmarino:
by the way, coordinating on the distfiles is *urgent*.
I was going to start building packages on pkgbox, but I can't merge the "staged" branch because it contains the bad rust ports along with a good pypy fix.  So the request to revert this has caused a jam, so I really need to know where the "good" distfiles are now.

jrmarino:
We lost 24 hours on this reversion request, which is time pkgbox64 should have been building.  I'm promoting the current branch right now which has needed updates to pypy and speech-dispatcher and the staged versions of rust is what is getting built as well.

mneumann:
@jrmarino: I asked you if it's an issue to reuse the same names for exactly that reason. I would have liked to have changed the path from "2017-04-27" to something else, to distinguish from the old snapshots, but this would not have been possible with further patches to the rust sources :(. You said, it's ok to reuse the same names again, as we didn't had published yet. 

The bootstrap files that are available right now, are the old ones, which should work! But you need to revert this patch (old sha256 hashes) to make it work. Sorry if this caused a lot of trouble. But I noticed too late that the bug with cargo and multiple threads still exists. I wrongly assumed it was related to our pthread fix, which it wasn't. Now I think it is related to a multi-threaded cargo which is forking... and randomly hanging due to fork()+threads is not playing nice together.

jrmarino:
maybe I didn't understand the question.
In general, every iteration of bootstrap files should have unique names, even if it means additional patches to accommodate them.

mneumann:
ok, makes absolutely sense.

jrmarino:
The 4.9 packages are published now.  latest versions rust and firefox are now availing in package form.

PR #789, opened 2017-07-24
devel/cmake: For a rainy day.

zrj-rimwis:


jrmarino:
cmake 3.9.0 is out (it's only a matter of time before it hits freebsd ports, and it's already in ravenports).  It's possible these patches won't apply to the next release of cmake.

zrj-rimwis:
oh ports cmake is really stale, so will wait for updated one

zrj-rimwis:
Just line differences, patches apply cleanly.

jrmarino:
let's hope it still applies to 3.8.2 because I don't know how long it will take freebsd to upgrade.  they probably have to execute exp_runs.

PR #791, opened 2017-07-25
devel/libpci: Make basic functionality work without root privileges on DragonFly.

ivadasz:
Wit this patch, libpci will fall back to opening /dev/pci read-only, if the
user can't open it read-write, and it will then use the PCIOCGETCONF and
PCIOCGETBAR ioctls to retrieve information, instead of doing "raw" PCIOCREAD
and PCIOCWRITE ioctls.

This is needed to make the libpci usage in the chromium browser work
correctly.

### DIFF
--- a/ports/devel/libpci/dragonfly/patch-lib_fbsd-device.c
+++ b/ports/devel/libpci/dragonfly/patch-lib_fbsd-device.c
@@ -1,0 +1,293 @@
+diff --git a/lib/fbsd-device.c b/lib/fbsd-device.c
+index ab04954..78067dc 100644
+--- lib/fbsd-device.c
++++ lib/fbsd-device.c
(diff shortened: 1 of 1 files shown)


jrmarino:
this doesn't appear that it would apply if I merged it.
the patch contents have:
```
+--- a/lib/fbsd-device.c
 ++++ b/lib/fbsd-device.c
```

That would require stripping the first component off, which is a rare setting.  The  /devel/libpci/Makefile doesn't have the setting.  Are you saying this patch needs to be regenerated to work?

ivadasz:
You're right, sorry. I've stripped the first component of the path now. Also I renamed the patch file to corretly begin with "patch-", instead of "patch_" which wouldn't be applied automatically.

PR #796, opened 2017-09-26
databases/redis-devel: added support of kqueue, fixed build with lua

yagelix:
Fixed build when redis is  configured to use lua or luajit from dports.
Added checks for DragonFly to use kqueue multiplexing API instead of select.


jrmarino:
FYI, the patches had the wrong path in them.
It seems this PR wasn't tested.  Please use synth in the future to verify proposed changes actually compile, thanks.

PR #797, opened 2017-09-29
Fix a couple of sysctls so psutil can be imported and does not have

peeterm007:
runtime issues.

* spyder (version 3.x) runs now fine.


peeterm007:
Sorry I started from the wrong end. Can you see the changes now?

PR #798, opened 2017-09-30
Update to wayland-1.13.

peeterm007:
* The patches in files are essentially the same as before, just adjusted
  to wayland-1.13.

I ran weston-1.9 with an uptime of one week without any problems with hardware accelerated rendering. Firefox was fine too (via Xwayland).

It was the first time I could run weston with hardware accelerated rendering, and it worked fine. For this I built mesa-libs with the 'wayland' option enabled, and built weston with the gl-renderer.


jrmarino:
This wasn't tested in synth or poudriere, was it?

```

Package 'libxml-2.0', required by 'virtual:world', not found

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.
```

That error implies it was built on a live system only.

peeterm007:
That's correct. Apologies. Added missing dependencies and built with synth: success. I sent a patch via new pull request

https://github.com/DragonFlyBSD/DeltaPorts/pull/801

PR #800, opened 2017-10-05
www/chromium: Replace with newport, and update to 60.0.3112.116.

ivadasz:
Do I have to do anything with the STATUS file, or is it going to be automatically updated accordingly for the switch to a newport?


ivadasz:
Changed the first line of the STATUS file to DPORT, to indicate that it is a newport now.

PR #814, opened 2017-12-01
net/shadowsocks-libev: Remove DragonFly-specific Makefile

liweitianux:
The DragonFly-specific Makefile only disables the ``-Werror`` compiler flag, but it is no longer needed.

I've tested with the latest ``net/shadowsocks-libev`` (v3.1.0 from FreeBSD Ports) on DragonFly-5.1.

### DIFF
--- a/ports/net/shadowsocks-libev/Makefile.DragonFly
+++ b/ports/net/shadowsocks-libev/Makefile.DragonFly
@@ -1,2 +1,0 @@
-dfly-patch:
-	${REINPLACE_CMD} -e "s|-Werror ||" ${WRKSRC}/src/Makefile.in


jrmarino:
well, does the dfly-patch target still succeed in removing "-Werror"?
If so, it should stay.  No port should be built with Werror, ever ...
or in other words, is there a problem just leaving this in place?

liweitianux:
@jrmarino , thank you for the reply.

> well, does the dfly-patch target still succeed in removing "-Werror"?

The current ``dfly-patch`` only takes care of the ``src/Makefile``, but there are still other places with ``-Werror`` enabled, e.g., ``configure.ac``, ``{libcork,libipset}/CMakeLists.txt``.

> No port should be built with Werror, ever ...

Sorry, but I don't get the point here.  If the developers and porters managed to enable the ``-Werror`` compiler flag, wouldn't it be better?

> is there a problem just leaving this in place?

No, this port (including the new version) builds fine with this ``Makefile.DragonFly``.

I just want to propose a cleanup to reduce the efforts to maintain.

jrmarino:
no, a port is never better with -Werror.
-Werror should never be used on a release build.  On a developer/debug build, sure.
This is a long standing debate.  So people feel the opposite that I do (but they're wrong!).
basically, new warnings are added by compiler developers all the time.  Over time, perfectly good code will eventually emit a warning as compilers get newer.  The build should never break just by upgrading the compiler.  -Werror causes all sorts of problems for ports/pkg collections.

by the way, it's possible for -Werror to be in configure scripts but not be propagated to the actual build...

liweitianux:
@jrmarino , thank you for the insightful explanation.

I'll just cancel this Pull Request, and try to propose your idea (i.e., ``-Werror`` should be disabled on release builds) to the upstream when I can figure out a patch to the build framework.

jrmarino:
well, we don't need to proactively disable Werror.  However, once Werror breaks a build, we do the work to find Werror and remove/disable it, so we might as well keep it.

jrmarino:
(meaning there are tons of ports that have -Werror set and it's too much work to hunt them all down)

liweitianux:
> (meaning there are tons of ports that have -Werror set and it's too much work to hunt them all down)

Yeah. No one can get everything work perfectly.  I just interested to something I could to make the softwares that I use into better shapes.

Thank you @jrmarino :-)

jrmarino:
well, without removing this file, the port doesn't build.  No makefile.in file exists, so patching fails.

PR #818, opened 2017-12-27
This fixes the problem with graphics/inkscape not finding lib/inkscape_base.so

peeterm007:
This fixes the problem with graphics/inkscape not finding lib/inkscape_base.so

The problem is that although inkscape is linked with rpath, the
linking options are wrong. In addition to

-Wl,-rpath,"$ORIGIN/../lib/inkscape:/usr/local/lib"

they should also contain -Wl,-z,origin:

-Wl,-rpath,"$ORIGIN/../lib/inkscape:/usr/local/lib" -Wl,-z,origin

setting MAKE_ENV+= LDVER=ld.bfd adds the required flag.

Thanks to zrj for pointing out this.

Background: similar discussion on FreeBSD and libreoffice. It does not
apply any more though since FreeBSD does not care of whether FLAGS, FLAGS_1 is set.

https://bugs.documentfoundation.org/show_bug.cgi?id=54015

### DIFF
--- a/ports/dns/powerdns/STATUS
+++ b/ports/dns/powerdns/STATUS
@@ -1,3 +1,3 @@
 PORT
-Last attempt: 4.0.4_6
-Last success: 4.0.4_6
+Last attempt: 4.1.0_1
+Last success: 4.1.0_1


jrmarino:
I'm missing the information where gold linker can't handle $ORIGIN.
I don't think that's true.   I don't understand why ld.bfd is doing something ld.gold isn't.

jrmarino:
in other words, "-Wl,-z,origin" is not specific to ld.bfd.  That's just the flag that is required to activate $ORIGN in the rpath.

peeterm007:
Yes, that's true. I was too cryptic, let me elaborate.

The root of the issue is that DragonFly behaves differently from both FreeBSD and Linux with respect to how $ORIGIN is handled. DragonFly needs "-Wl,-z,origin" at linking so that "(FLAGS) ORIGIN" and "(FLAGS_1)  Flags: ORIGIN" are set in the binary. If the latter are set, $ORIGIN is expanded correctly; if not, $ORIGIN is not expanded and the shared library is not found. The requirement that the latter two flags are needed in order for $ORIGIN expansion seems standard behavior; or at least it was standard at some point.

However, FreeBSD and Linux do not care if "(FLAGS) ORIGIN" and "(FLAGS_1)  Flags: ORIGIN" are set in the binary. I.e. even if they are not set, $ORIGIN in rpath or runpath is still expanded correctly and the shared library is found. I tested this with a small example found in the thread above

https://bugs.documentfoundation.org/show_bug.cgi?id=54015

This is the reason inkscape fails to start on DragonFly but not on FreeBSD. I installed inkscape on FreeBSD and tested that it runs fine. The inkscape header on FreeBSD reads: 

% readelf -d /usr/local/bin/inkscape | grep ORIGIN
.... RPATH  Library rpath: [$ORIGIN/../lib/inkscape:/usr/local/lib]
.... RUNPATH  Library runpath: [$ORIGIN/../lib/inkscape:/usr/local/lib]

I do not know why FreeBSD and Linux behave so.

It is correct that ld.bfd and ld.gold do not differ with respect to "-Wl,-z,origin". Both need this explicitly in order that "(FLAGS)" etc be in the binary on DragonFly. The difference is that with "MAKE_ENV+= LDVER=ld.bfd", the required "-Wl,-z,origin" is added in the linking phase. Question: why are those flags added with ld.bfd?

(As a side, compiling with "-Wl,-z,origin" on FreeBSD results in a binary where "(FLAGS)" and "(FLAGS_1)" are just empty. Compiling with "-Wl,-z,origin,-rpath" as suggested in the libreoffice thread above yields empty fields too. No idea how to correctly set "(FLAGS)" etc on FreeBSD. But then it does not matter, FreeBSD does not care anyway.)

All in all it might not be a good idea to fix this using LDVER=ld.bfd. One could have used LDFLAGS+= "-Wl,-z,origin", this explicitly shows the fix. The result is the same, the "(FLAGS)" etc are set correctly and inkscape runs fine.

In light of this, there are at least three options: 

(1) USE_LDCONFIG+= ${PREFIX}/lib/inkscape,
(2) LDFLAGS+= "-Wl,-z,origin",
(3) LDVER=ld.bfd

What's the best option to use? 

The general question is whether or not DragonFly might want to align the behavior with respect to $ORIGIN with Linux/FreeBSD?

jrmarino:
option 2 is the best.
I'm guessing the reason z,origin isn't requires on linux, freebsd (or solaris) is because they use ld.bfd and dragonfly uses gold linker by default, thus gold linker must explicitly require the flag (which is debatably incorrect).

But if just adding -Wl,-z,origin to LDFLAGS fixes the issue, that's the right solution.

peeterm007:
OK I'll prepare a patch with (2) tomorrow.

PR #819, opened 2017-12-29
Fix the problem with graphics/inkscape not finding lib/inkscape_base.so

peeterm007:
The issue is fixed by adding "-Wl,-z,origin" to LDFLAGS.


peeterm007:
Just as a followup to the discussion in the previous PR: both ld.bfd and ld.gold on DragonFly need this flag to be specified explicitly in order that the required fields "(FLAGS) ORIGIN" and "(FLAGS_1) Flags: ORIGIN" are set in the binary.

PR #822, opened 2018-02-22
devel/rubygem-childprocess - Fix support for DragonFly BSD

tuxillo:
- This fix makes it possible to run vagrant on dfly
- If we ever get https://github.com/enkessler/childprocess/pull/135 upstream this
  patch can be removed.


tuxillo:
PR enkessler/childprocess#135 was merged, we might need to revert that commit on the next update

PR #823, opened 2018-02-23
sysutils/packer - Fix build

tuxillo:
This port is untested and it will potentially fail to work
properly but it's a start ;-)


tuxillo:
It requires DISABLE_LICENSES=yes to build, not sure what's wrong but the way the Makefile is defined is not accepted in DragonFly's Dports

PR #835, opened 2018-04-18
emulators/qemu - Fix build

tuxillo:
- Please note in order to build emulators/qemu, net/vde2 is required and this
  only builds on latest master/release branches where the BIOCFEEDBACK ioctls
  have been added.


jrmarino:
so qemu is not expected to build on DragonFly 5.0 or 5.2 releases with this change?

tuxillo:
emulators/qemu will build on the next 5.2 minor release because the necessary changes to get net/vde2 have been MFC'ed.
It won't in 5.0.x unless we MFC all the libpcap changes including the import etc.

PR #844, opened 2018-05-08
special: Add handling for gcc80.

zrj-rimwis:
While there, add extra hook for late cmake args.


jrmarino:
1) is c++17 not supported?
2) I'm still concerned about fortran mismatch.  I think gcc8 fortran should be used instead of gcc5 fortran (except apparently lang/gcc8 doesn't exist yet, wtf?)

PR #847, opened 2018-05-10
devel/openmp: Decouple this lomp lib.

zrj-rimwis:
We use gomp OpenMP implementation. Also avoid mismathing versions of lomp.


jrmarino:
I assume math/tmw is really math/tmv.  will fix.

zrj-rimwis:
yeah, math/tmv

PR #860, opened 2018-05-27
www/qt5-webengine: Add chromium patches.

zrj-rimwis:
For the next synced bulk. QT5 version was bumped. There will be issue w/ qt5-network and libressl. Will look how to add patch from HardenedBSD.

While there, unblock misc/qt5-doc, it should no longer fail (only repacks now).
In turn do the same for qt5-examples, since distfile no longer blocks.

Submitted-by: dillon


jrmarino:
doesn't build:
/construction/www/qt5-webengine/qtwebengine-opensource-src-5.7.1/src/3rdparty/chromium/build/build_config.h:61:2: error: #error Please add support for your platform in build/build_config.h

PR #869, opened 2018-09-19
Bring in Rust 1.29.0

mneumann:


jrmarino:
this never got pulled in because STATUS is set to LOCK.   It would be really helpful to unlock the port, otherwise it's going to be like it was never added.  I just got burned again with firefox.

jrmarino:
we need rust 1.30.1 now, asap.
FreeBSD ports tree already set dependencies on it.  
I will try to bump them back to 1.29, and hopefully mozilla products can still use 1.29.
If not, the lack of rust 1.30.x is blocking the latest dports sync.

PR #881, opened 2018-11-26
Update wayland to 1.16.

peeterm007:
* Note that wayland now provides libwayland-egl.so, which is also
  provided by mesa-libs when option WAYLAND is on. The latter option
  is not turned on by default in mesa-libs, so there is no conflict
  with the default mesa-libs settings. It will be probably removed
  soon from the mesa-libs.


jrmarino:
thanks!

PR #882, opened 2018-12-05
Change the location of input.h.

peeterm007:
* Starting with DragonFly 5.4.0, we export input.h to userland at

  /usr/include/compat/linux

  in the long run, this will reduce the number patches for both dports and ravenports.

### DIFF
--- a/ports/x11-toolkits/gtk30/dragonfly/patch-gdk_wayland_gdkdevice-wayland.c
+++ b/ports/x11-toolkits/gtk30/dragonfly/patch-gdk_wayland_gdkdevice-wayland.c
@@ -9,7 +9,7 @@
  #include <sys/time.h>
  #include <sys/mman.h>
 -#include <linux/input.h>
-+#include <dev/misc/evdev/input.h>
++#include <compat/linux/input.h>
  
  #define BUTTON_BASE (BTN_LEFT - 1) /* Used to translate to 1-indexed buttons */
  


peeterm007:
Yes sure it's a good idea. If I understand correctly, we support two releases + master in dports? Then we have two cases atm, (1) version => 5.4 (5.4 and master) and (2) version < 5.4 (essentially 5.2). 

If we need to support 5.0 as well, then we need to include a third case, where we have to rely on input.h from the dport multimedia/v4l_compat.

Question: are two cases (1), (2) above enough, or do we need to include 5.0 too?

jrmarino:
for dports, 2 cases are enough.
For ravenports, we'd probably need all three (and maintain dependency on video4linux even if it's not used for a while).

peeterm007:
The general idea is that we want to decouple our input.h from FreeBSD since we can only use the input.h that our kernel agrees with. So the choice's are as follows: 

(1) use our input.h etc that we have exported at /usr/include/compat/linux. This entails we have to patch ports with sth like

-I/usr/include/compat

in order that includes of the form 

#include <linux/input.h>

work.

(2) Or we can fake, i.e. make our own copy of evdev-proto whose content (input.h, input-event-codes.h) will be taken from our tree. Then we do not need patches like the ones in (1) on top of the ones that FreeBSD already have for input.h

(3) Or we go with FreeBSD's evdev-proto. This will lead to many headaches down the road since our kernel interface might not be in sync with what FreeBSD/Linux have.


Option (1) was the initial one we came up with. But now I realize that perhaps (2) is also possible, and at least at first sight needs less patching than (1). 

One issue that will come up (but that will come up with all approaches) is dependency problem. Port A may require that libinput be not less than X.Y, but our lower stack (our input.h) might yet not support that.

peeterm007:
Just to add that in case of (2) our own version of evdev-proto would be even simpler than FreeBSD's since we do not need any replacements. We'd just copy two files 

/usr/include/compat/linux/input.h
/usr/include/compat/linux/input-event-codes.h

to 

/usr/local/include/linux

What FreeBSD are doing is downloading Linux's versions of those files and copying them to /usr/local/include/linux

We do not have uinput.h since we do not have the corresponding functionality in the kernel. It could be ported from FreeBSD if necessary.

ftigeot:
We could make things even simpler by having the files directly available in /usr/include .
Is there an advantage of copying files to /usr/local/include/linux instead of having them in /usr/include/linux or adding /usr/compat/include/linux to the compiler search path ?

peeterm007:
Actually, the original idea was to have 

/usr/include/linux

but then swildner cautioned against this because applications might make unwarranted assumptions by looking whether 

/usr/include/linux/input.h

exists or not. For instance, it may suddenly happen that a number of applications start to build since they find /usr/include/linux/input.h and happily assume things we do not have in place yet. I agree with this; the more cautious approach where we do create the structure linux/input.h, but can switch it on by including the path say

/usr/include/compat

gives us control over what will happen. This is option (1) above. This is in fact what we have in place now: we have exported input.h at /usr/include/compat/linux, we'd just need to add -I/usr/include/compat where necessary.

peeterm007:
And on top of this I should mention that evdev is not in the default kernel config. One needs to rebuild kernel with 

options EVDEV_SUPPORT
device		evdev

to get that. I very much hope it'll be in the default kernel config in the next release. Right now it does no matter much since all (relevant) applications still work w/o evdev.

PR #907, opened 2019-05-14
Fix the tools in devel/libevdev

peeterm007:
* I missed a few events in the last patch. The
  mouse tool now works fine. The touchpad tool
  I could not test, but at least the mainloop
  is fine.


zrj-rimwis:
Thx

PR #918, opened 2019-11-20
[NEW Mk/Uses/dlang.mk] Initial commit of USES=dlang support script

dkgroot:
Feature:        d compiler
Usage:  USES=dlang
Valid args:     bootstrap       use bootstrap compiler when no valid d-compiler could be found

Additional variables that can be provided:
for any:
- DLANG_USES=
        additional features enabled in dlang module
        - dub
        - rdmd
        not implemented yet (added when need arises):
        - (reggae)
        - (cmake-d / cmake)
        - (COOK)
        - (BAZEL)
        - (BUTTON)
        - (MESON)
- DCOMPILERS
        List of d-compilers to search for through (in order)(default: ldc gdc dmd)
- USE_LDC=yes | USE_GDC=yes | USE_DMD=yes
        Overwrite auto selection
- DRT_MINVER
        Minimal required d-runtime version (inclusive) (default:20791)
- DRT_MAXVER
        Maximum allowed d-runtime version (inclusive) (default:20999)
- DLANG_DEBUG
        Print extra debug output for USES=dlang

for ldc:
- D_REQ_LLVM    LLVM version to require (60 70 80)

for gdc:
- D_REQ_GCC     GCC version to require (8 9 10)

Variables provided by dlang.mk (after bsd.ports.mk has been loaded)
for any:
- CHOSEN_DCOMPILER_TYPE
        D compiler selected
- DCOMPILER
        Full path to D-compiler selected.
- DCOMPILER_DMD
        Full path to DMD interface compatible compiler selected.
- DRUNTIME_VERSION
        D runtime version of selected D-compiler.
- DFLAGS
        Addional DFLAGS variables to be passed to the D-compiler.
for ldc:
- LLVM_CONFIG
        Full path to llvm-config being used
- LLVM_VER
        Major version of llvm being used
for gdc:
- GCC_VER
        Major version of gcc being used

MAINTAINER: ddegroot@talon.nl

Signed-off-by: ddegroot <ddegroot@talon.nl>

PR #920, opened 2019-11-21
[NEW PORT] lang/gdc : initial port for gcc-9.2.0 d-compiler

dkgroot:
Can use either use lang/dmd2 or lang/ldc or fall back to bootstrapping using gcc
Added gdmd script for dmd2 reference compatibility

dependencies:
- Mk/Uses/dlang.mk (separate PR #918)

MAINTAINER: ddegroot@talon.nl
Signed-off-by: ddegroot <ddegroot@talon.nl>


tuxillo:
What do we do with this one?

daftaupe:
> Can use either use lang/dmd2 or lang/ldc or fall back to bootstrapping using gcc

I see there was an attempt to port lang/ldc as a newport, but now it has fallback to using fports land/ldc and doesn't build yet, what is supposed to happen on that subject ?
@dkgroot are you willing to maintain the whole dlang ecosystem ? Meaning lang/gdc, the other port submission lang/dm2 and maybe lang/ldc and the diff needed for DragonFly support ?

dkgroot:
@daftaupe @tuxillo
I tried to push several Makefiles to FreeBSD for gdc, ldc and dmd2 quite a while ago (2019), and made very little to no progress in getting those changes accepted, making it hard to land them here as well. Getting the USES=dlang makefiles merged will be even hard then a regular port apparently.
I originally created the dragonflybsd ports for all three projects making it possible for d-lang to run on dragonflybsd and was expecting that to be the hard part. As a changed jobs in the meantime, it might be a bit of a challenge picking all of this up again. The PR's provided should still work, but might require a little TLC. It's up to you what you wanna do, drop or merge.

daftaupe:
Alright @dkgroot, thanks for the feedback. I guess we'll have to discuss this with @tuxillo.

PR #931, opened 2020-05-14
Fix EGL Window creation and prevent GLFW-Wayland from SEGFAULTING

FrostKiwi:
These are fixes that I needed to apply in order for glfwCreateWindow() to stop SEGFAULTING, with the main culprit being EXT_image_dma_buf usage, despite not being supported. 
The patches allow window creation on sway, with Mesa 18.3.2 and newer. However, due to EXT_KHR not being properly used, the window remains blank with Mesa 18.3.2, even though a context is successfully created. Proper EXT_KHR extension use was fixed at some point and with Mesa 20.2.0-devel (git-412e29c277) at the latest, window and context creation works without issue.

All patch files were diffed against Mesa 20.2.0-devel (git-412e29c277).

### DIFF
--- a/ports/graphics/mesa-dri/dragonfly/patch-src_egl_drivers_dri2_egl_dri2.c
+++ b/ports/graphics/mesa-dri/dragonfly/patch-src_egl_drivers_dri2_egl_dri2.c
@@ -1,0 +1,15 @@
+diff --git a/src/egl/drivers/dri2/egl_dri2.c b/src/egl/drivers/dri2/egl_dri2.c
+index 2e51d28a8b9..95ee2f3a7b1 100644
+--- a/src/egl/drivers/dri2/egl_dri2.c
++++ b/src/egl/drivers/dri2/egl_dri2.c
+@@ -1005,8 +1005,8 @@ dri2_setup_screen(_EGLDisplay *disp)
+ #ifdef HAVE_LIBDRM
+       if (dri2_dpy->image->base.version >= 8 &&
+           dri2_dpy->image->createImageFromDmaBufs) {
+-         disp->Extensions.EXT_image_dma_buf_import = EGL_TRUE;
+-         disp->Extensions.EXT_image_dma_buf_import_modifiers = EGL_TRUE;
++         disp->Extensions.EXT_image_dma_buf_import = EGL_FALSE;
++         disp->Extensions.EXT_image_dma_buf_import_modifiers = EGL_FALSE;
+       }
+ #endif
+    }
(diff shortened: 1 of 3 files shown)


daftaupe:
@FrostKiwi sorry for not taking care of this PR earlier, but would you still have a working DragonFly setup running wayland/Sway ?
We would like to test the current state of wayland before trying to upgrade it to the version of the FreeBSD ports (see the draft in https://github.com/DragonFlyBSD/DeltaPorts/pull/1293)

FrostKiwi:
> @FrostKiwi sorry for not taking care of this PR earlier, but would you still have a working DragonFly setup running wayland/Sway ? We would like to test the current state of wayland before trying to upgrade it to the version of the FreeBSD ports (see the draft in #1293)

Ahh, sure. I actually switched to FreeBSD, but I'll throw DragonFlyBSD again on a machine an check the state of Window creation.

tuxillo:
> > @FrostKiwi sorry for not taking care of this PR earlier, but would you still have a working DragonFly setup running wayland/Sway ? We would like to test the current state of wayland before trying to upgrade it to the version of the FreeBSD ports (see the draft in #1293)
> 
> Sure. I actually switched to FreeBSD, but I'll throw DragonFlyBSD again on a machine an check the state of Window creation.

If you could actually describe the steps maybe we can do it ourselves. Let me know :)

FrostKiwi:
> > > @FrostKiwi sorry for not taking care of this PR earlier, but would you still have a working DragonFly setup running wayland/Sway ? We would like to test the current state of wayland before trying to upgrade it to the version of the FreeBSD ports (see the draft in #1293)
> > 
> > 
> > Sure. I actually switched to FreeBSD, but I'll throw DragonFlyBSD again on a machine an check the state of Window creation.
> 
> If you could actually describe the steps maybe we can do it ourselves. Let me know :)

Ohh sry. Kind of stuck with other tasks to reinstall a whole OS to fix something that might not be there anymore.

The crux of the issue was Wayland window creation segfaulting under specific situations, because DragonFlyBSD did not support `EXT_image_dma_buf`, which caused a black window without context to appear under the specific combination of a Wayland compositor running a wayland window created via GLFW.

Unless I misremember something, to confirm this still being an issue would be to create a Window using GLFW in Wayland mode under swaywm as the compositor. GLFW usually comes default in X11 mode and has to  be recompiled with `GLFW_USE_WAYLAND` set to true. Then window creation failed on DragonFlyBSD, which was traced to `EXT_image_dma_buf` missing with the help from the Mesa dev's IRC channel.

PR #941, opened 2020-07-09
databases/rocksdb-lite : fix package

daftaupe:
For some reason the build of the lib version was also trying to package the non-lite files in the lib dir (which are not built in that case, I wonder why that doesn't break on FreeBSD side), so I had to manually exclude them with the %%NOLITE%% conditionnal when building -lite variant.

These files are the following :
lib/librocksdb.a
lib/librocksdb.so
lib/librocksdb.so.6
lib/librocksdb.so.%%SHLIB_VER%%
lib/librocksdb.so.%%PORTVERSION%%
lib/librocksdb_tools.a

I did try to rebuild both rocksdb and rocksdb-lite after the change.

### DIFF
--- a/ports/databases/rocksdb/diffs/Makefile.diff
+++ b/ports/databases/rocksdb/diffs/Makefile.diff
@@ -1,6 +1,6 @@
---- Makefile.orig	2019-11-02 19:40:49 UTC
-+++ Makefile
-@@ -14,7 +14,6 @@ LICENSE_COMB=	dual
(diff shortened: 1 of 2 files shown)


tuxillo:
I think we should investigate why it is not the case for FreeBSD.

daftaupe:
@tuxillo it has been fixed in https://github.com/freebsd/freebsd-ports/commit/85dd2f2ac2fa8b0140ec3248125cd1567cf0cc56 it seems.

PR #942, opened 2020-07-09
security/libp11 : fix package

daftaupe:


tuxillo:
why is the plist different for us?

PR #951, opened 2020-07-13
x11-toolkits/p5-Alien-wxWidgets : fix package

daftaupe:
Possibly something is still missing to get a perl module named correctly "gtk_3_0_5_uni_gcc_3_4.pm" for FreeBSD vs "gtk_3_0_5_uni_nc_0.pm" in our case.

I've checked, so far that package didn't even ship a file. Let's say it's a step forward :)


zrj-rimwis:
It was cc --version parsing issue, closing.

PR #967, opened 2020-07-21
lang/sbcl : fix build

daftaupe:
@tuxillo, from https://github.com/jrmarino/ravensource/blob/63c53237fb7b07b8ec69df48e826b697f440ed1a/bucket_FA/sbcl/specification

I notice that :

OPT_ON[dragonfly]=	DRAGONFLY
OPT_ON[freebsd]=	FREEBSD CONCURRENCY

and 

[CONCURRENCY].CMAKE_ARGS_OFF=		--without-sb-thread
					--without-restore-fs-segment-register-from-tls
[CONCURRENCY].CMAKE_ARGS_ON=		--with-sb-thread
					--with-restore-fs-segment-register-from-tls

So basically CONCURRENCY is the same set of options as what THREAD is in our Makefile, so I put back the THREADS_VARS to what it was and just disable the THREAD options for us.

I let you try, but it's building for me :)

### DIFF
--- a/ports/lang/sbcl/diffs/Makefile.diff
+++ b/ports/lang/sbcl/diffs/Makefile.diff
@@ -1,15 +1,15 @@
---- Makefile.orig	2020-06-28 18:44:21 UTC
-+++ Makefile
-@@ -97,7 +97,7 @@ SAFEPOINT_IMPLIES=	THREADS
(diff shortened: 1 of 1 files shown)


tuxillo:
As spoken, we might want to not disable threads entirely and check with the sbcl people what could be wrong.

daftaupe:
sbcl seems to build fine, so this PR is now useless

PR #977, opened 2020-08-10
x11-toolkits/wlroots : fix patch

daftaupe:


daftaupe:
Fixed by @zrj-rimwis so closing.

PR #987, opened 2020-08-20
devel/libphonenumber : fix patch

daftaupe:


zrj-rimwis:
These are publicly installed headers. We prefer if these would be patched even if USES= alias pass. Helps once such headers get used in other ports. Closing, will regen patches in delta.

PR #1121, opened 2021-09-01
Fix: misc/mc to compile with option ncurses

ttt-fifo:
Include search path fixed in order mc to compile with option
ncurses.


tuxillo:
Should be fixed in 4df3c9328f5abbec6c72fed163eb627ad53f5112, thanks for reporting!

PR #1143, opened 2021-11-10
sysutils/e2fsprogs : remove patch to fix check-plist phase

daftaupe:


tuxillo:
Are those binaries no longer present?

```
-/sbin/e2fsck
-/sbin/fsck_ext2fs
```

Or did they change the path?

daftaupe:
Actually, this commit removes a pkg-plist patch that was not taking them into account.

The message on sting was 
> ===> Checking for items in STAGEDIR missing from pkg-plist  
> Error: Orphaned: /sbin/e2fsck  
> Error: Orphaned: /sbin/fsck_ext2fs`

And as these files were in reality present
> % ls -l /build/buildbase/SL00/construction/sysutils/e2fsprogs/stage/sbin/
> total 1544
> -r-xr-xr-x  1 root  wheel  778536 Nov 11 13:30 e2fsck
> -r-xr-xr-x  1 root  wheel    7480 Nov 11 13:30 fsck_ext2fs`

We're now taking them into account again.

tuxillo:
Ah ok, excellent then, thanks.

PR #1162, opened 2021-12-14
Fix net/gerbera

liweitianux:
* Gerbera requires C++17 filesystem feature, which is not available in
  the base GCC 8.0.  Pull in the 'misc/cxxfs_gcc8' port to fix this
  requirement.

* Switch the default UPnP library from libnpupnp to libupnp, because
  libnpupnp (e.g., v4.1.5) fails to listen socket on DragonFly and thus
  needs further fix; libupnp works well.


zrj-rimwis:
Thanks, merged

PR #1186, opened 2022-08-08
Create symbolic links libopensmtpd.so.0 ...

mneumann:
This is required for `opensmptd-filter-dkim` to work properly, which wants `libopensmtpd.so.0`.


tuxillo (changes requested):
Can we use a variable instead of hardcoding the so version of the library? See SOVERSION usage across ports/dports.

mneumann:
@tuxillo second commit should address your comment!

tuxillo:
Looks good to me.

PR #1206, opened 2022-09-13
Fix `service h2o configtest`

mneumann:
- Running `service h2o configtest` fails as it does not find Perl.

- Running `/usr/local/etc/rc.d/h2o configtest` usually works as it has /usr/local/bin in PATH.

- FreeBSD is not affected as their /usr/sbin/service binary starts the h2o script with `env -L -/daemon`, which populates the env with those defined in /etc/login.conf for login class `daemon`, which happens to include /usr/local/bin in the PATH. This works, but is not really their intention, as /usr/local/etc/rc.d/h2o explicitly defines `h2o_perl` and should pass this as H2O_PERL environment variable!


mneumann:
Not sure if putting this under `diffs` is the right approach....

mneumann:
Ideally this would be upstreamed to FreeBSD ports :)

PR #1244, opened 2022-10-12
[new port] devel/efivar-bsd

SergiiDmytruk:
There is already `efivar` from FreeBSD and `libefivar` in the base system, but this one:
 * is of the latest version (38 instead of 0.15 (upstream changed versioning))
 * uses native kernel's API (similar to `libefivar` from the base system)
 * provides the same API as on Linux, allowing for porting apps that require it
 * provides `libefiboot` and `libefisec` (these won't work with `libefivar` from the base system)

Amount of changes is quite substantial and is very similar across *BSD systems, so a fork is used instead of a set of patches.

Fully implementing `libefiboot` for a system will take a lot of work, so it has limitations ([details](https://github.com/3mdeb/efivar/blob/bsd/BSD_PORTING.md)). Generating abbreviated device paths works and that's enough to boot an EFI payload, making it useful for `fwupd`.

This can also replace `efivar` port instead of adding a new one if you don't mind major library version change `0 -> 1`. Otherwise, need to mark `efivar` as conflicting with this one, can a patch be applied to FreeBSD port's `Makefile`?


tuxillo:
It was in the queue, I've forced it now:

https://github.com/DragonFlyBSD/DeltaPorts/commit/aa68663b752006715f001b1c3a0d7b559d53877d

SergiiDmytruk:
Nice, thanks.

PR #1256, opened 2022-10-15
x11-servers/xorg-server : fix patch phase

daftaupe:


daftaupe:
Please have a careful look at the second commit as it's about x11-servers/xorg-server@xorg, so it might have a rather significant impact.

PR #1296, opened 2022-11-05
x11-servers/xwayland : fix configure phase

daftaupe:


daftaupe:
FreeBSD has libmd in base, this makes meson finds automatically a way to get sha1 function.

PR #1375, opened 2023-05-29
sysutils/dmidecode: Emulate FreeBSD to support UEFI-only systems

liweitianux:
On UEFI-only systems, the SMBIOS Entry Point structure must be located by the SMBIOS_TABLE_GUID configuration table, which is now exported via the 'hint.smbios.0.mem' following the FreeBSD's way.

Patch sysutils/dmidecode to use the kenv.

See also: https://github.com/DragonFlyBSD/DragonFlyBSD/commit/5e488df32cb01056a5b714a522e51c69ab7b4612

### DIFF
--- a/ports/sysutils/dmidecode/Makefile.DragonFly
+++ b/ports/sysutils/dmidecode/Makefile.DragonFly
@@ -1,0 +1,1 @@
+USES+=	alias


liweitianux:
The patch has been accepted by the upstream :smile:

See https://savannah.nongnu.org/bugs/index.php?64261

daftaupe:
@liweitianux any idea about which version will provide your patch ?

liweitianux:
> @liweitianux any idea about which version will provide your patch ?

I'm sorry that I missed this.  The next upstream version (i.e., 3.6) will include my patch so the patch here should be reverted then.  This PR works around the current 3.5 version for us.

liweitianux:
> should we merge this?

Please merge this, so we'll get a working port for us.

We can revert this patch when the new version arrives and causes the patch failing to apply.

(Sorry for the delayed response.)

PR #1387, opened 2023-08-02
multimedia/gstreamer1-libav : disable tests target as it requires gst…

daftaupe:
…reamer-check, disabled in multimedia/gstreamer1 because of POSIX timers not available


tuxillo:
I think I overlapped this one and didn't see your PR, can it be closed?

daftaupe:
sure :)

PR #1388, opened 2023-08-02
net/geoclue : make use of Clang

daftaupe:


tuxillo:
I think it was also built successffully with gcc:

```
sting# ls -l /build4/boomdata/dsynth/packages/All/geoclue-2.5.7.pkg 
-rw-r--r--  1 root  wheel  248071 Aug  2 16:04 /build4/boomdata/dsynth/packages/All/geoclue-2.5.7.pkg
```

PR #1389, opened 2023-08-02
devel/libgusb : make use of Clang

daftaupe:


tuxillo:
I think it got built just fine this time without clang:

image

daftaupe:
You're right, so closing this one :)

PR #1402, opened 2023-09-08
www/privoxy: Fix segfault error upon action editting in Web UI

liweitianux:
This patch has also been submitted to the upstream: https://sourceforge.net/p/ijbswa/bugs/940/


liweitianux:
Upstream has accepted the fix and should appear in the next version (i.e., v3.0.35).

tuxillo:
should we merge this then?

liweitianux:
> should we merge this then?

Please merge this, so we'll get a working port for us.

We can revert this patch when the new version arrives and causes the patch failing to apply.

(Sorry for the delayed response.)

PR #1464, opened 2023-12-06
Add newport `mkisofs` - stripped down `cdrtools`

mneumann:
Using `sysutils/mkisofs` instead of `sysutils/cdrtools` mainly gets rid of various sound-specific dependencies (LAME, VORBIS). This idea is to use this "port" also in nrelease and dillons `withpkgs` branch.


mneumann:
@tuxillo any idea why the CI build fails? what am I doing wrong?

liweitianux:
It's the INDEX generation that's currently failing, so no worries ;)

On the other hand, I don't think you should add the `STATUS` file; it will be auto added by the builder, if i understand correctly.

tuxillo:
The STATUS file is needed because it is a DPORT, we have the full definition in DeltaPorts.

mneumann:
@tuxillo if you think this port is worth (maybe under a different name), feel free to merge. it should drop quite a few dependencies. Alternative would be to create a "flavor" of cdrtools.

tuxillo:
> @tuxillo if you think this port is worth (maybe under a different name), feel free to merge. it should drop quite a few dependencies. Alternative would be to create a "flavor" of cdrtools.

hmmm, maybe creating a flavor would be actually better?

mneumann:
@tuxillo: Flavor would work, though it'd be a bit more "complex" to implement and it would make it more difficult to keep in sync with FreeBSD, unless they adopt the same patches.

daftaupe:
@mneumann from what I understand, the port is going to evolve together when cdrtools / schilybase are updated, so we don't have to sync it manually everytime they change version ?

If that's the case, I'm ok with that port, it makes sense to not bring all the dependencies in the base image.

mneumann:
@daftaupe yes, it includes Makefile.master from schilybase. So, it will evolve like cdrtools when schilybase gets an update. it might need "tweaking" iff new binaries would be added to schilybase, as we remove them in post-install. Though that is highly unlikely...

mneumann:
@daftaupe yeah, the dependency list is huge. lots of sound-related stuff.

liweitianux:
Well, I've just migrated our nrelease framework to replace `mkisofs` with in-base `makefs(8)` utility; see https://github.com/DragonFlyBSD/DragonFlyBSD/commit/57b363e86aca95e0a7308e855326c76528ba2bc7 :smiley:

mneumann:
@liweitianux that's awesome, thanks! I think, in this case, we can ignore my port :).

PR #1465, opened 2023-12-16
net/zerotier: Fix building and cleanup existing patches

liweitianux:
- Add one patch (`patch-ext_prometheus-cpp-lite-1.0_core_include_prometheus_text__serializer.h` to fix the building of 1.12.2 version.
- Regenerate the existing patches as a result of `make makepatch`.
- Remove unnecessary diffs to `files/zerotier.in`.
- Restore the FreeBSD's `ZEROTIER_SYSLOG` substitution.


liweitianux:
See also my previous PR to ZeroTier: https://github.com/zerotier/ZeroTierOne/pull/1043

PR #1470, opened 2024-03-06
net/minidlna: Ignore sparseness check to make 'rescan' work

liweitianux:
The sparseness check function `check_notsparse()` isn't supported by DragonFly, and it caused the 'rescan' to not work; i.e., updated files were not indexed upon startup when the `-r` option was specified.

Just ignore the sparseness check to make the 'rescan' work as expected.

By the way, the `-r` option is recommended to add for minidlna; for example, add this to `/etc/rc.conf`:

```
minidlna_enable="YES"
minidlna_flags="-r"
```


liweitianux:
Tested with the latest minidlna 1.3.3 from FreeBSD ports.

PR #1472, opened 2024-06-02
Update www/h2o port to match latest FreeBSD port

mneumann:
* This PR gets rid of www/h2o-devel and updates www/h2o.

* Our www/h2o port is EOL since end of last year.

* In FreeBSD, www/h2o-devel is gone, long live www/h2o.


mneumann:
@tuxillo can I merge that one?

tuxillo:
We're still on 2024Q1 and i see `www/h2o-devel` still exists, why remove it?

PR #1473, opened 2024-06-14
Depend on base64 which FreeBSD has in their base

mneumann:


daftaupe:
lgtm, will merge once I'm sure @tuxillo is not building stuff

PR #1475, opened 2024-08-11
Updating devel/tree-sitter to match latest FreeBSD port

fac3plant:
This PR upgrades devel/tree-sitter to match the latest FreeBSD port. This port is required for editors/neovim 10.1 (matching latest FreeBSD port) to build.

### DIFF
--- a/ports/devel/tree-sitter/STATUS
+++ b/ports/devel/tree-sitter/STATUS
@@ -1,3 +1,3 @@
 PORT
-Last attempt: 0.20.8
-Last success: 0.20.8
+Last attempt: 0.22.6
+Last success: 0.22.6
--- a/ports/devel/tree-sitter/diffs/Makefile.diff
+++ b/ports/devel/tree-sitter/diffs/Makefile.diff
@@ -1,0 +1,12 @@
+diff --git a/devel/tree-sitter/Makefile b/devel/tree-sitter/Makefile
+index c751e4a17e1..a0b51645e27 100644
+--- a/devel/tree-sitter/Makefile
++++ b/devel/tree-sitter/Makefile
+@@ -1,6 +1,6 @@
+ PORTNAME=	tree-sitter
+ DISTVERSIONPREFIX=	v
+-DISTVERSION=	0.20.8
++DISTVERSION=	0.22.6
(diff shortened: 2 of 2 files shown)


daftaupe:
Hi @fac3plant,

thanks for your effort.
Unfortunately that's not really how it works (maybe our doc has to be modified ?).

First of all, the STATUS file.
If the first line is PORT, with Last attempt and Last success, that means we're following the FreeBSD port. In that case, we don't try another version than what is in the targetted QX branch of the moment (in that case we're following 2024Q1, so we're using https://github.com/freebsd/freebsd-ports/blob/f78a53fb71d1755da714dfc20ee75829c59c3652/devel/tree-sitter/Makefile#L3, which references the version 0.20.8). We will reach 0.22.6 when targetting 2024Q3 as specified by https://github.com/freebsd/freebsd-ports/blob/c6199cc1de2f4a14ec12dc5b4b5ab13421ae0e5e/devel/tree-sitter/Makefile#L3

One addition is that the STATUS file is not to be modified by hand, it's done automagically when the official builder succeeds (@tuxillo operates it).

Last attempt: -> that line represents which version comes from FreeBSD ports (last one we tried)
Last success: -> that version is the same as the first one if the builder built the version attempted properly, otherwise the version is lower and that means that the port has to be fixed for some reasons.
If you check https://github.com/DragonFlyBSD/DeltaPorts/blob/master/ports/devel/tree-sitter/STATUS, you'll see both line with 0.20.8, that means that the port has built properly for the target version 0.20.8 (doesn't mean that the port works as expected, but that's another story).

So as you might have guessed now, one would not patch the Makefile to reach the proper version as it will be tried automatically when we start following that branch. But then some fixes might be needed at that time.

Sometimes you can notice a DPORT line before the 2 others, instead of PORT (as in https://github.com/DragonFlyBSD/DeltaPorts/blob/a5f2768de49af0ab486d88cad13df3c308d70d21/ports/net/libradius/STATUS#L1), that means that port doesn't follow a FreeBSD port. In that case everything is contained in the directory called newport.

I have to close this PR, but don't hesitate to submit more fixes and ask if things are still not clear.

PR #1477, opened 2024-08-21
Sysutils/screen Add session creation time when list active sessions 

fac3plant:
Reimplement [this](https://github.com/freebsd/freebsd-ports/commit/7418bfa0e65838e7f94bd82d1e02c5aac79b0555) patch from FreeBSD to be able to see a session's creation time when running `screen -ls`.


daftaupe:
@fac3plant I have only looked at the patch real quick, but what about the other files (manpage for example) ? Does your patch applies on top of the FreeBSD one ? or is it standalone ?

daftaupe:
@fac3plant looks good, I get the following in my chroot (applying your patch on top of the staged version) : 
```
build:/usr/dports/sysutils/screen % doas chroot /build/dsynth/build/SL00
# screen -ls
There are screens on:
	530161.pts-2.build	(23/08/2024 19:04:54)	(Attached)
	530153.pts-0.build	(23/08/2024 19:04:32)	(Attached)
2 Sockets in /tmp/screens/S-root.

```
On comment though, the LDFLAGS modification is already in the original Makefile of the port on line `60 LDFLAGS+=       -L${NCURSESLIB}`
So I think only the -lkvm is needed on that line.

fac3plant:
@daftaupe I believe I have corrected Makefile.DraognFly.

fac3plant:
@daftaupe The latest commit removes the diffs directory, and should allow the other patches to apply without issue. I was able to use them to build sysutils/screen on the staged branch in dports without issue. However, please let me know if something needs to be corrected.

daftaupe:
@fac3plant thanks !

PR #1480, opened 2024-09-15
Updating editors/neovim to match FreeBSD's 2024Q3 neovim

fac3plant:
This patch updates editors/neovim to match the 2024Q3 FreeBSD Ports editors/neovim.


tuxillo:
The STATUS file needs to be there, not sure why it got deleted.

fac3plant:
I saw a different PR looked like it had deleted the contents of the STATUS. I can restore it can update this PR.

PR #1482, opened 2024-10-04
Fixing math/calc build failures.

fac3plant:
This patch adds a 'DragonFly' build target which allows math/calc to build properly.

### DIFF
--- a/ports/math/calc/dragonfly/patch-Makefile.target
+++ b/ports/math/calc/dragonfly/patch-Makefile.target
@@ -1,0 +1,52 @@
+--- Makefile.target.orig	Fri Oct  4 06:12:04 2024
++++ Makefile.target	Fri Oct
+@@ -255,6 +255,49 @@ MAKE= gmake
+ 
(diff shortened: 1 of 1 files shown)


ports/math/calc/dragonfly/patch-Makefile.target
```diff
@@ -0,0 +1,52 @@
...
++CCMISC=
++LCC= gcc
++CC= ${PURIFY} ${LCC} ${CCWERR}
++
++# We must use gmake as the FreeBSD target because under some older
++# releases of FreeBSD, make not support conditional syntax.
++#
++MAKE= gmake
++
++endif	# ($(target),FreeBSD)
```
daftaupe, line 49: wouldn't it make more sense to have `endif	# ($(target),DragonFly)` here ?

fac3plant: It would make more sense to do that; I just didn't see it. Thank you for catching it, and that correction has been made.

daftaupe:
Thanks !

PR #1484, opened 2024-11-12
Adding new port www/CryptPad.

fac3plant:
This commit adds CryptPad as a port to DragonFlyBSD's DPorts tree as well as adding a user and group for the service to run.


daftaupe:
@fac3plant any reason not to propose this as a FreeBSD port and then just profit from it in the DPorts ?

fac3plant:
I have proposed this as a FreeBSD [port](https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=280486), and if/when it gets accepted there, I think it would be better to profit from that. I haven't heard anything back regarding the port since I opened in on 2024-07-29, so I thought I would submit it for dports, then transition if/when it gets submitted on FreeBSD ports. If that is not a good plan, then it is understandable.

daftaupe:
@tuxillo what's your view on this one ?

IMO the best way is to wait for this port to enter FreeBSD's tree.

tuxillo:
Yes please, let's wait for it. We're almost finishing 2024Q3 now.

PR #1485, opened 2024-12-16
Updating port to match FreeBSD ports from 2024Q3 branch.

fac3plant:
This patch simply removes the dragonfly patches that are no longer necessary as 7-zip has DragonFlyBSD as a target in the code. The port files from FreeBSD ports 2024Q3 work on DragonFly without modification.


daftaupe:
thanks !

PR #1497, opened 2025-03-21
devel/xsd: Fix man page locations

kworr:


ports/devel/xsd/Makefile.diff
```diff
@@ -0,0 +1,13 @@
+diff --git a/devel/xsd/Makefile b/devel/xsd/Makefile
```
tuxillo, line 1: The location for diffs is in `diffs` within the port, not in the port directory itself, see other ports as an example.

PR #1501, opened 2025-05-07
security/ca_root_nss: Patch pkg-plist to not run certctl

liweitianux:
certctl is a in-base utility in FreeBSD that manages certificates and trust stores.  We don't have it and I don't think we current need to port it over, because we don't yet have /etc/openssl/certs and other trust stores.

See also: https://lists.dragonflybsd.org/pipermail/users/2025-May/452480.html


daftaupe:
Thanks @liweitianux !

PR #1503, opened 2025-06-10
x11-servers/xorg-server: Re-enable glamor hw accel, disabled by default

servizig:
Intel i915 in master branch now supports hardware acceleration, bring back "AccelMethod" option to Xorg server, disabled by default for backwards compatibility.


daftaupe:
Thanks for both PRs @servizig !

PR #1506, opened 2025-08-03
devel/libffi: Enable FFI_MMAP_EXEC_WRIT

servizig:
This is a temporary patch until we get this fix from upstream.


daftaupe:
Thanks !
@servizig @liweitianux I was actually wondering if that one should be backported to the master branch of dports ?

liweitianux:
> Thanks ! @servizig @liweitianux I was actually wondering if that one should be backported to the master branch of dports ?

Yah, please do that.

PR #1532, opened 2026-03-11
htop: replace man1 with share/man1 in pkg plist

autumnjolitz:


autumnjolitz:
Any questions about this PR?

daftaupe:
Hello, from what I can see, in 2026Q1 branch of FreeBSD ports, that change is already in https://cgit.freebsd.org/ports/tree/sysutils/htop/Makefile?h=2026Q1#n20

daftaupe:
Just checked, and can confirm that with htop 3.4.1, the man page is in share/man/man1
`SL00/construction/sysutils/htop/stage/usr/local/share/man/man1/htop.1.gz`

autumnjolitz:
Is this released yet? Because from my experience, here’s what happened:

1. I tried installing a rather common utility named `htop`
2. I found it wasn’t available on `pkg`
3. I take **my** time to setup dports using the available documentation which *supposedly* has best practices in it
4. I find the issue and make a PR
5. Weeks later I’m told the issue is already fixed in a *magical unavailable not released*/**doesn’t exist** version
6. I and others are still left with broken `htop` because no one is bothering to push said magically fixed dports downstream for anyone else to use

I am well aware that volunteer time is limited. That is why I spent my time to try to make it better for other people.

This experience has left me rather drained, disappointed and disheartened. It has also been rather disempowering.

daftaupe:
@autumnjolitz I answered you while having not much time, so I surely didn't give enough details in my answers. I'm sorry if the whole experience fell so bad for you.

From the details you gave, I think something in our doc is not clear enough : DeltaPorts is the repo we use to build the to-be-released DPorts (the current result arrives in the staged branch of the DPorts repo). That's why I'm talking about what you call "magical unavailable not released/doesn’t exist version", it's the version we're working on, in this case 2026Q1 branch of the FreeBSD ports.

From DeltaPorts perspective (and next batch of ports we will release), your patch is not required, as it's already correct in the FreeBSD ports tree we are using, 2026Q1 branch. But in the currently shipped binary packages (based on DPorts master branch), it's helpful, so as we are in the Delta repo, I answered following the first perspective.

If you want to fix the master branch of DPorts (what you get when getting ports through pkg's official mirrors, which is based on FreeBSD ports 2025Q3 at the moment I'm writing), you can fix by opening a PR directly in the DPorts repo, working on the master branch.

Thanks again for contributing, I hope we can call that a false start and count you as a DPort contributor.
You can come to IRC or write emails if you have any question.

